PR #27, opened 2015-02-19
Remove DEFAULT_NUM_CONSUMER_THREADS that's no longer used

michaelklishin:


michaelklishin:
For @simonmacmullen to QA.

PR #28, opened 2015-02-19
Clarify that ConnectionFactory#setRequestedHeartbeat sets a timeout

michaelklishin:
...and not communication interval

Also, link to our upcoming doc guide.


michaelklishin:
For @simonmacmullen to QA.

Issue #42, opened 2015-04-22
AMQConnection missing "/" for non-default virtual hosts


dharrigan:
Hi,

In AMQConnection, line 865, if the virtualHost, is *not* the default, the resultant toString looks something like this:

```
amqp://guest@192.168.80.3:5672test
````

It should be

```
amqp://guest@192.168.80.3:5672/test
```

PR coming shortly.

-=david=-

michaelklishin:
Fixed by #44. Thank you, @dharrigan!

dharrigan:
np :)

dumbbell:
Since RabbitMQ 1.4.0.

PR #43, opened 2015-04-22
Issue #42: AMQConnection missing "/" for non-default virtual hosts.

dharrigan:
If the virtual host is not the default virtual host, then there needs
to be a separator between the port and the name of the virtual host.

-=david=-


michaelklishin:
@dharrigan thank you. May I ask you to re-submit this against the `stable` branch? This can go into a bug fix release.

michaelklishin:
Thank you.

Closing in favour of #44, will merge stable into master.

PR #45, opened 2015-04-23
Fix deadlock on close() with congested server.

dpratt:
The call BlockingRPCContinuation#getReply used in the close() method of ChannelN does not specify a timeout. If the server is congested, this call to close() can (and does) block indefinitely and eats up the current thread. 

From an API point of view, a close() method should really have a timeout associated with it, and if the server doesn't respond with a cleanly closed connection, the channel should just be discarded.


michaelklishin:
@dpratt sounds reasonable. Can you please re-submit this against the `stable` branch?

dpratt:
Done.

PR #71, opened 2015-06-24
Remove WorkPool memory leak

### LINKED ISSUE #69: WorkPool possible memory leak in Set<K> unlimited
Short story: `unregisterKey(K key)` and `unregisterAllKeys()` does not unregister the key from `Set<K> unlimited`, which seems to result in leaks over time.

Longer story:

In our (Spring) application, we dynamically spin up and spin down queues and `SimpleMessageListenerContainer`s as needed. I can describe the workflow in more detail if anyone really cares. After a day or two of decent load, our memory usage tends to grow very large. Analyzing a heap dump showed that the majority of memory was being used by a few hundred thousand `ShutdownSignalException`s and `ChannelN`s and various other related objects that still had references, all of which led back to this one `Set` in the `WorkPool` class.

It seems like a simple solution would be to add two lines which also call `unlimited.remove`, and I can create a Pull Request if desired, but I'm hoping that someone more familiar with the code can determine if this would cause any problems.

Thanks for all you folks do on this project. It's been fun implementing it into my system.

Link to feasible patch: https://gist.github.com/xathien/ec9ca35224868a524873

xathien:
Fixes #69

### DIFF
--- a/src/com/rabbitmq/client/impl/WorkPool.java
+++ b/src/com/rabbitmq/client/impl/WorkPool.java
@@ -123,6 +123,7 @@ public class WorkPool<K, W> {
             this.pool.remove(key);
             this.ready.remove(key);
             this.inProgress.remove(key);
+            this.unlimited.remove(key);
         }
     }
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Thank you! It will be in `3.5.4`.

PR #76, opened 2015-07-21
Cannot Close AutorecoveringConnection While Disconnected From Network

dhakimTRX:
I took a look at your changes to the dotnet client for handling close/abort while disconnect from network and added something similar to the java client.  

Added manuallyClosed flag to AutorecoveringConnection which prevents recovery of connections after application code has closed/aborted the connection.

I've run it through tests in my application, Android no longer leaks threads named AMQP Connection <ip>:<port> when closing while WiFi is disabled.

### DIFF
--- a/src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
+++ b/src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
@@ -68,11 +68,11 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
     private final List<ConsumerRecoveryListener> consumerRecoveryListeners = new ArrayList<ConsumerRecoveryListener>();
     private final List<QueueRecoveryListener> queueRecoveryListeners = new ArrayList<QueueRecoveryListener>();
 	
-	//Used to block connection recovery attempts after close() is invoked.
+	// Used to block connection recovery attempts after close() is invoked.
 	private volatile boolean manuallyClosed = false;
 	
-	//This lock guards the manuallyClosed flag and the delegate connection.  Guarding these two ensures that a new connection can never
-	//be created after application code has initiated shutdown.  
+	// This lock guards the manuallyClosed flag and the delegate connection.  Guarding these two ensures that a new connection can never
+	// be created after application code has initiated shutdown.  
 	private Object recoveryLock = new Object();
 	
     public AutorecoveringConnection(ConnectionParams params, FrameHandlerFactory f, Address[] addrs) {
@@ -188,8 +188,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      * @see com.rabbitmq.client.Connection#close()
      */
     public void close() throws IOException {
-		synchronized(recoveryLock)
-		{
+		synchronized(recoveryLock) {
 			this.manuallyClosed = true;
 		}
         delegate.close();
@@ -199,8 +198,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      * @see Connection#close(int)
      */
     public void close(int timeout) throws IOException {
-		synchronized(recoveryLock)
-		{
+		synchronized(recoveryLock) {
 			this.manuallyClosed = true;
 		}
         delegate.close(timeout);
@@ -210,8 +208,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      * @see Connection#close(int, String, int)
      */
     public void close(int closeCode, String closeMessage, int timeout) throws IOException {
-		synchronized(recoveryLock)
-		{
+		synchronized(recoveryLock) {
 			this.manuallyClosed = true;
 		}
         delegate.close(closeCode, closeMessage, timeout);
@@ -221,8 +218,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      * @see com.rabbitmq.client.Connection#abort()
      */
     public void abort() {
-		synchronized(recoveryLock)
-		{
+		synchronized(recoveryLock) {
 			this.manuallyClosed = true;
 		}
         delegate.abort();
@@ -232,8 +228,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      * @see Connection#abort(int, String, int)
      */
     public void abort(int closeCode, String closeMessage, int timeout) {
-		synchronized(recoveryLock)
-		{
+		synchronized(recoveryLock) {
 			this.manuallyClosed = true;
 		}
         delegate.abort(closeCode, closeMessage, timeout);
@@ -243,8 +238,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      * @see Connection#abort(int, String)
      */
     public void abort(int closeCode, String closeMessage) {
-		synchronized(recoveryLock)
-		{
+		synchronized(recoveryLock) {
 			this.manuallyClosed = true;
 		}
         delegate.abort(closeCode, closeMessage);
@@ -254,8 +248,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      * @see Connection#abort(int)
      */
     public void abort(int timeout) {
-		synchronized(recoveryLock)
-		{
+		synchronized(recoveryLock) {
 			this.manuallyClosed = true;
 		}
         delegate.abort(timeout);
@@ -296,8 +289,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      * @see com.rabbitmq.client.Connection#close(int, String)
      */
     public void close(int closeCode, String closeMessage) throws IOException {
-		synchronized(recoveryLock)
-		{
+		synchronized(recoveryLock) {
 			this.manuallyClosed = true;
 		}
 		delegate.close(closeCode, closeMessage);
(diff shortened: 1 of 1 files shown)


src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -402,18 +441,21 @@ public void removeConsumerRecoveryListener(ConsumerRecoveryListener listener) {
...
+		this.recoverShutdownListeners();
+		this.recoverBlockedListeners();
+		this.recoverChannels();
+		if(this.params.isTopologyRecoveryEnabled()) {
+			this.recoverEntities();
+			this.recoverConsumers();
+		}
 
-        this.notifyRecoveryListeners();
+		this.notifyRecoveryListeners();
```
michaelklishin: Please pay attention to trailing whitespace.

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -428,18 +470,43 @@ private void recoverBlockedListeners() {
         }
     }
 
-    private void recoverConnection() throws IOException, InterruptedException {
-        boolean recovering = true;
-        while (recovering) {
+	//Returns true if the connection was recovered, 
+	//false if application initiated shutdown while attempting recovery.  
+    private boolean recoverConnection() throws IOException, InterruptedException {
```
michaelklishin: I disagree with the decision to turn a side-effecting method into a boolean. It should be a separate method (or even a condition on `manuallyClosed`).

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -428,18 +470,43 @@ private void recoverBlockedListeners() {
...
+						return true;
+					}
+				}
+
+				//This is the once in a blue moon case.  
+				//Application code just called close as the connection
+				//was being re-established.  So we attempt to close the newly created connection.
+				try
+				{
+					newConn.close(); //Is there a standard error code/shutdown message to use here?
```
michaelklishin: If you want to close the delegate and ignore all exceptions, use `Connection#abort()`.

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -67,7 +67,14 @@
...
     private final List<ConsumerRecoveryListener> consumerRecoveryListeners = new ArrayList<ConsumerRecoveryListener>();
     private final List<QueueRecoveryListener> queueRecoveryListeners = new ArrayList<QueueRecoveryListener>();
-
+	
+	//Used to block connection recovery attempts after close() is invoked.
+	private volatile boolean manuallyClosed = false;
+	
+	//This lock guards the manuallyClosed flag and the delegate connection.  Guarding these two ensures that a new connection can never
+	//be created after application code has initiated shutdown.  
+	private Object recoveryLock = new Object();
```
michaelklishin: Synchronising on `this` is not discouraged in Java.

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -428,18 +470,43 @@ private void recoverBlockedListeners() {
...
+				//This is the once in a blue moon case.  
+				//Application code just called close as the connection
+				//was being re-established.  So we attempt to close the newly created connection.
+				try
+				{
+					newConn.close(); //Is there a standard error code/shutdown message to use here? 
+				}
+				catch (Exception e)
+				{
+					//Ugh, we just leaked the new connection object, hopefully some kind of heartbeat/expiration timeout will take care of it.
```
michaelklishin: If the connection is kept alive, there is no expiration and heartbeats, in fact, will *keep it open* (and the object will have references). Please do not leave comments you are not 100% sure about, as incorrect comments
are worse than not having any.

Use `Connection#abort` as recommended above.

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -428,18 +470,43 @@ private void recoverBlockedListeners() {
         }
     }
 
-    private void recoverConnection() throws IOException, InterruptedException {
-        boolean recovering = true;
-        while (recovering) {
+	//Returns true if the connection was recovered, 
+	//false if application initiated shutdown while attempting recovery.  
+    private boolean recoverConnection() throws IOException, InterruptedException {
+        while (!manuallyClosed) //manuallyClosed marked volatile -ensures thread cannot cache value of manuallyClosed.
```
michaelklishin: If the intent of using `while` here is to bypass thread-local caching, this is redundant since the field is volatile. If the intent is to double check field value, this is quite unclear. I'd rather move connection initiation inside the synchronised block (we want to only ever have 1 delegate per autorecovering connection anyway).

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -428,18 +470,43 @@ private void recoverBlockedListeners() {
...
+				{
+					if (!manuallyClosed)
+					{
+						//This is the standard case.
+						this.delegate = newConn;					
+						return true;
+					}
+				}
+
+				//This is the once in a blue moon case.
```
michaelklishin: Comments in the rest of the codebase begin with a space.

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -402,18 +441,21 @@ public void removeConsumerRecoveryListener(ConsumerRecoveryListener listener) {
...
-        this.recoverConnection();
-        this.recoverShutdownListeners();
-        this.recoverBlockedListeners();
-        this.recoverChannels();
-        if(this.params.isTopologyRecoveryEnabled()) {
-            this.recoverEntities();
-            this.recoverConsumers();
-        }
+        if (!this.recoverConnection())
+			return; //Application initiated a shutdown during automatic recovery, don't attempt further recovery
```
michaelklishin: Comments in the rest of the codebase are on the line above.

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -181,6 +188,10 @@ public boolean isOpen() {
      * @see com.rabbitmq.client.Connection#close()
      */
     public void close() throws IOException {
+		synchronized(recoveryLock)
```
michaelklishin: I'm pretty sure this indentation is not what the rest of the codebase (or even this particular class) uses.

michaelklishin:
Please make changes per the comments above and re-submit this against the `stable` branch, then we can include it into `3.5.5`. Thank you!

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
dhakimTRX: beginAutomaticRecovery() was marked synchronized, and holds that lock
throughout the connection recovery process.  I would not be able to
synchronize on the AutorecoveringConnection object from within the
close/abort methods without locking up the application code.  That's why I
created the secondary lock object.

On Tue, Jul 21, 2015 at 5:59 PM, Michael Klishin <notifications@github.com>
wrote:

> In src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
> <https://github.com/rabbitmq/rabbitmq-java-client/pull/76#discussion_r35159924>
> :
>
> > @@ -67,7 +67,14 @@
> >      private final Map<String, RecordedConsumer> consumers = new ConcurrentHashMap<String, RecordedConsumer>();
> >      private final List<ConsumerRecoveryListener> consumerRecoveryListeners = new ArrayList<ConsumerRecoveryListener>();
> >      private final List<QueueRecoveryListener> queueRecoveryListeners = new ArrayList<QueueRecoveryListener>();
> > -
> > +	
> > +	//Used to block connection recovery attempts after close() is invoked.
> > +	private volatile boolean manuallyClosed = false;
> > +	
> > +	//This lock guards the manuallyClosed flag and the delegate connection.  Guarding these two ensures that a new connection can never
> > +	//be created after application code has initiated shutdown.
> > +	private Object recoveryLock = new Object();
>
> Synchronising on this is not discouraged in Java.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rabbitmq/rabbitmq-java-client/pull/76/files#r35159924>
> .
>



-- 
 *  Daniel Hakim*
  TRX Systems Inc.
  dhakim@trxsystems.com
  Phone: +1 (301) 313-0053 Ext 731

Developer of NEON™ Personnel Location Products
www.trxsystems.com

 The information contained in this e-mail and any attachment is covered by
the Electronic Communications Privacy Act, 18 U.S.C. 2510-2521, may contain
TRX Systems, Inc. proprietary and/or confidential information, and is
intended only for the confidential use of the above named recipient. Please
contact the sender if you receive this email and it is not intended for you.

Follow TRX Systems: <http://feeds.feedburner.com/trxsystems/hXJN>
<http://www.linkedin.com/companies/932098> <http://twitter.com/TRX_SYSTEMS>

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
dhakimTRX: The intention was a double check of the manuallyClosed flag.  

It was my assumption that the connection initiation was network io with a timeout.  If that is the case, I don't want to be doing it while holding a lock, as that would cause close/abort calls to unexpectedly block application code.  If that's not a worry, (which maybe it shouldn't be, as a successful close call appears to do network io as well), then I can move it into the synchronized block and remove the double check.

src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
dhakimTRX: I don't see another way to do this without restructuring beginAutomaticRecovery.  I cannot check manuallyClosed after calling recoverConnection() without a potential race condition in which I leak the connection object.  beginAutomaticRecovery needs to know if a new connection was created.

michaelklishin:
Closing in favour of #77.

PR #77, opened 2015-07-22
Stops Recovery When Client Closes

dhakimTRX:
Adds a boolean flag, manuallyClosed to keep track of whether or not the
AutoRecoveringConnection has been told to close by the application.

If flag is set before recovery attempt, autorecovery fails out.
If flag is set mid recovery attempt, creates the new
connection, then immediately aborts it.

Comments, whitespace match with approved style.  

Did not address issue of recoverConnection returning boolean, could not see way to do so without introducing race condition or requiring larger scope of changes.

Did not modify double checked locking in recoverConnection, am not sure of policies in this library for what public methods are allowed to block.


michaelklishin:
Thank you!

michaelklishin:
Thank you!

PR #80, opened 2015-08-22
Add travis-ci configuration

gmr:
This is a minimal configuration that only adds ``test-client``.

I'd like to enhance it to support more but with the current structure, it'll prove a bit difficult. I'm thinking given Travis's support for docker, it makes sense to make a RabbitMQ image specifically for testing that has the SSL config setup since Travis's RabbitMQ service doesn't appear to support SSL.

Additionally, the requirement of accessing RabbitMQ via rabbitmqctl makes me think having a testing docker image makes sense as well, as we would have more control over the service/process than the travis service level abstraction.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+sudo: false
+language: erlang
+addons:
+  apt:
+    packages:
+      - xsltproc
+services:
+  - rabbitmq
+otp_release:
+  - "R16B03-1"
+install:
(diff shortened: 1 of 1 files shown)


michaelklishin:
@gmr since Java client also tests the server, I'd make it run `make lite` from `rabbitmq-test`. It would take longer but would generate TLS certificates and run TLS as well as multi-nodes. How does that sound?

michaelklishin:
Will switch it to use `make lite` from `rabbitmq-test`. Thank you!

PR #88, opened 2015-09-18
Use an executor for shutdown monitors

### LINKED ISSUE #87: Use an executor for shutdown threads
See #86.

michaelklishin:
Fixes #87, references #86.


YannRobert:
@michaelklishin There is a remaining place where the ```shutdownExecutor``` should be used :  ```com.rabbitmq.client.impl.ChannelManager#scheduleShutdownProcessing``` (this is the place I spotted in my original issue)

michaelklishin:
Filed https://github.com/rabbitmq/rabbitmq-java-client/issues/94 for that.

PR #95, opened 2015-09-22
Allow to set an optional Executor to be used to send heartbeats

YannRobert:
In a consistent fashion with the ability to set an Executor for shutdown and for ```ConsumerWorkService```, 
allow to set an optional ```ScheduledExecutorService``` to be used to send heartbeats.

If the ```ScheduledExecutorService``` is not provided to the ```ConnectionFactory```, one ```ScheduledExecutorService``` per Connection (and one Thread per Connection) will be created at runtime (as it was before).
If the ```ScheduledExecutorService``` is provided to the ```ConnectionFactory```, the provided ```ScheduledExecutorService``` will be used.

This feature can be useful for the following reasons : 
- Because sending a HeartBeat is generally really fast, it may be a waste to have 1 Thread per Connection in the JVM. HeartBeats can support to be a few millseconds/seconds late. Using a shared ```ScheduledExecutorService``` with a very small amount of Threads for all of the Connections allows to use less Threads, with no significant drawbacks.
- Threads will be named by the Executor, and are easier to find in a Thread Dump or a Profiler

### DIFF
--- a/src/com/rabbitmq/client/ConnectionFactory.java
+++ b/src/com/rabbitmq/client/ConnectionFactory.java
@@ -20,14 +20,11 @@ import java.io.IOException;
 import java.security.KeyManagementException;
 import java.security.NoSuchAlgorithmException;
 import java.util.Map;
-import java.util.concurrent.ExecutorService;
+import java.util.concurrent.*;
 
 import java.net.URI;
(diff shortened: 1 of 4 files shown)


michaelklishin:
Thank you, looking into it as we speak.

PR #98, opened 2015-09-24
Better use Executor#execute than ExecutorService#submit

YannRobert:
see https://github.com/rabbitmq/rabbitmq-java-client/commit/30a9603a9fb9aaf02b518670e4b1c19e1ac8953c#commitcomment-13395844


michaelklishin:
@YannRobert thank you!

Issue #120, opened 2015-12-27
channel.queueDelete sometimes throws NullPointerException, possibly due to race on recordedBindings


spatula75:
When multiple channels share an AutorecoveringConnection and queues are coming and going from multiple theads, about 1 time out of 10000 , we are seeing a NullPointerException or ConcurrentModificationException to be thrown:

```
java.lang.NullPointerException
        at com.rabbitmq.client.impl.recovery.AutorecoveringConnection.removeBindingsWithDestination(AutorecoveringConnection.java:768)
        at com.rabbitmq.client.impl.recovery.AutorecoveringConnection.deleteRecordedQueue(AutorecoveringConnection.java:690)
        at com.rabbitmq.client.impl.recovery.AutorecoveringChannel.deleteRecordedQueue(AutorecoveringChannel.java:562)
        at com.rabbitmq.client.impl.recovery.AutorecoveringChannel.queueDelete(AutorecoveringChannel.java:303)
        at com.rabbitmq.client.impl.recovery.AutorecoveringChannel.queueDelete(AutorecoveringChannel.java:299)
```

The code at this location looks innocuous enough:
```
    Set<RecordedBinding> removeBindingsWithDestination(String s) {
        Set<RecordedBinding> result = new HashSet<RecordedBinding>();
        for (Iterator<RecordedBinding> it = this.recordedBindings.iterator(); it.hasNext(); ) {
            RecordedBinding b = it.next();
            if(b.getDestination().equals(s)) { // [[ NPE here ]] //
                it.remove();
                result.add(b);
            }
        }
        return result;
    }
```

However, `removeBindingsWithDestination` is not synchronized, so it's possible for multiple threads to be in this method changing the recordedBindings ArrayList at the same time as bindings come and go from multiple threads, even with channels not being shared across threads (but the Connection is shared).

It looks the other accesses to recordedBindings are synchronized (though caution: `maybeDeleteRecordedAutoDeleteExchange` is synchronizing differently than the others) so `removeBindingsWithDestination` may have just fallen through the cracks.

It may be a sufficient fix just to make `removeBindingsWithDestination` synchronized.

(Another tangential thing I noticed is that most of the recorded maintenance is done from synchronized methods, but `maybeDeleteRecordedAutoDeleteQueue` during a `basicCancel` instead synchronizes on `recordedQueues` and `consumers`, and that this method just removes the `recordedQueues` entry, not the `recordedBindings` associated with the queue as well.)

[...]

PR #121, opened 2015-12-30
rabbitmq-java-client #120: unsynchronized access to recordedBindings gives undesired results

spatula75:
Fix is to simply make `removeBindingsWithDestination` a synchronized method (like the other methods which access recordedBindings).

### DIFF
--- a/src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
+++ b/src/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
@@ -766,7 +766,7 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
         return result;
     }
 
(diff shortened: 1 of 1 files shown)


michaelklishin:
Other similar methods synchronise on the specific fields they mutate. Can we make this synchronise on `recordedBindings`?

spatula75:
I'm sure it's possible; however, all of the other methods which are currently synchronized would need to change to the field-synchronized strategy as well.  There's sort of a mix of different synchronization styles in play currently.

michaelklishin:
Ah, yes, there are `synchronized` methods in `AutorecoveringConnection`, too. Alright then.

spatula75:
Yeah, it might be slightly less-than-ideal, but probably a bigger project to revise how concurrency works in that particular class.

michaelklishin:
@spatula75 agreed. Thank you!

Issue #37, opened 2015-03-26
Migrate build system away from Ant


michaelklishin:
Ant has taken us quite far but it's showing its age. I feel we should follow what most modern (and very likely future) JVM-based projects at Pivotal use: Gradle. We'll be able to piggyback on the existing Spring team infrastructure and have nightly snapshots published to Bintray, Spring repositories. Releases also can then easily be published to the two mentioned above plus Maven Central.

Finally, Gradle is just a nicer and more widely used tool. Good idea for making it easier to contribute.

PatrickSauts:
The amqp-client already has a maven pom.xml file.
Gradle integration in eclipse is still not efficient even in GGTS.
Maven is still the most widely used.

My point is you might want to do both at least for dependency management for now.

What do you think ?

michaelklishin:
Maven is fine but most Java projects at Pivotal use Gradle.

I also find Maven to be a real pain when you need to do
something non-standard, as our test tasks do. Gradle is OK in this regard.

michaelklishin:
@PatrickSauts lets put it this way: we have ant targets that shell out (start external processes). If you can help us find a suitable and sane Maven plugin for that, I'd consider Maven (after consulting with the Spring team). Maven would be a significant step in the right direction for us.

PatrickSauts:
There is an https://maven.apache.org/plugins/maven-antrun-plugin/. I used in the past might fulfill the need.
I'm looking into it right now.
The deployment I believe should also be handled by maven instead of a sh script, right ? Any other concerns ?

michaelklishin:
@PatrickSauts it depends what exactly we mean by "deployment" :) I see no problem with having a shell script that uses Maven and basic *nix tools as appropriate.

Starting nodes in our test suite and standard JAR packaging/Bundlor are the hardest parts. @dumbbell may be able to suggest something else.

Just to set some expectations: this will be merged after 3.6.0 ships, and thus #114 will have to wait until 3.7. Better late than never, of course, and you can safely run a JAR from master with a released server version, the protocol has been very stable for the last 5 years.

michaelklishin:
@PatrickSauts thank you for stepping up to help, by the way!

PatrickSauts:
@michaelklishin My pleasure

PatrickSauts:
@michaelklishin 
[...]

PR #119, opened 2015-12-22
#37 Migrate build system away from Ant (WIP)

PatrickSauts:
Fixed #37 (at least to a large extent).

The tests can be run with `mvn clean test`
or with `ant`, I kept it compatible.

And also with the right `settings.xml` in your `~/.m2` directory like
```
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
                      http://maven.apache.org/xsd/settings-1.0.0.xsd">
<servers>
    <server>
      <id>sonatype-nexus-staging</id>
      <username>my_login</username>
      <password>my_password</password>
    </server>
  </servers>
</settings>
```
you can run `mvn clean deploy` that would compile, run the tests, and deploy to the repository if the tests pass.

I also added maven version plugin, so to set a version you can do it that way
`mvn versions:set -DnewVersion=<new.version> && mvn versions:commit`

That's a better alternative than the release plugin.
So basically to release a new version the maven command would be:
`mvn versions:set -DnewVersion=<new.version> && mvn versions:commit && mvn clean deploy`

PR #122, opened 2016-01-08
Move build system to Maven (retains Ant tasks for backwards compatibility)

michaelklishin:
Originally submitted by @PatrickSauts as #119.

This moves build system to Maven while retaining backwards compatibility by offering "proxy" Ant tasks.


dumbbell:
Fixes #37.

dumbbell:
The SSL testsuite does bot work because `isSSlAvailable()` crashes for me and, after when this is fixed, it returns false because `sslClientPath` points to a directory, not a file. I'm working on it.

PatrickSauts:
Super merci !

dumbbell:
A bit more details about my last commit: what was done in `AbstractRMQTestSuite.java` was in fact never called. Maybe because we still rely on JUnit 3 API, not 4. For example, the `runTest()` was never called so the properties initialized and verified in `AbstractRMQTestSuite.java` was ineffective.

This probably remained unnoticed because default values for `rabbitmqctl.bin` and `sibling.rabbitmq_test.dir` are ok when the Java client is inside the Umbrella. However, the default value for `make.bin` only works when GNU Make is installed as `make`, which is not the case on *BSD (GNU Make is available as `gmake`).

Our testsuite didn't handle well failures from `make`. Usually, this means the testsuite is stuck. I extended what @PatrickSauts did and added calls from the TestSuites themselves, to be sure we have all properties set before running tests.

Finally, the new `broker.{hostname,port,sslport}` properties were good when using Maven, but not Ant. That's what confused me yesterday. Now, the top-level `build.xml` explicitely sets those properties and `isSSLAvailable()` is back on rails.

dumbbell:
Hmm, there is one more thing to fix: `ClientVersion.java` is correctly generated by `pom.xml` using the `impl.version` property (set in the top-level `Makefile`), but `pom.xml` hard-codes the version for itself (currently set at `3.6.1-SNAPSHOT`).

dumbbell:
@PatrickSauts: Merged, thank you very much!

PatrickSauts:
@dumbbell 
When moving completely from ant 
Uncomment line 148 and remove line 147 in pom.xml


Have the right `settings.xml` in your `~/.m2` directory like
```
<settings xmlns="http://maven.apache.org/SETTINGS/1.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 xsi:schemaLocation="http://maven.apache.org/SETTINGS/1.0.0
<servers>
    <server>
      <id>sonatype-nexus-staging</id>
      <username>my_login</username>
      <password>my_password</password>
    </server>
  </servers>
</settings>
```

To release a new version the maven commands would be:
`mvn versions:set -DnewVersion=<new.version> && mvn versions:commit && mvn clean deploy`

It would set the version in the pom and use that version in the generated java file.


By the way you can also run the perf tests like this
`mvn -q exec:java -Dexec.classpathScope=test -Dexec.mainClass="com.rabbitmq.examples.PerfTest"`

or with arguments
```
mvn -q exec:java -Dexec.classpathScope=test -Dexec.mainClass="com.rabbitmq.examples.PerfTest" -Dexec.args="arg0 arg1 arg2" 
```

Basically any class with a main inside  package com.rabbitmq.examples without the need to moving any jar around, the classpath is defined by maven.

I added a `runperftestMaven.sh` and `stresspersisterMaven.sh` inside the script folder.

One last thing I can write an assembly to build a zip to run perf tests that would contain all the necessary files.

michaelklishin:
@PatrickSauts note that we also distribute the JAR directly from GitHub and rabbitmq.com. But

```
mvn versions:set -DnewVersion=<new.version> && mvn versions:commit
```

certainly sounds right up our alley, thank you!

PatrickSauts:
@michaelklishin You probably want to build the jars using maven too right ?

`mvn versions:set -DnewVersion=<new.version> && mvn versions:commit && mvn clean package`

I'm pretty sure that was implicit in your comment but just in case.

So moving ~completely away from ant you could get rid of `lib` and 'bundlor` folders, all bundlor files and build.xml.

Your very welcome. Glad I could help.

michaelklishin:
@PatrickSauts sure, thank you. We will ping you once we have more Maven questions (which should be around the time `3.7.0` Milestone 2 ships, so in about a month).

PR #127, opened 2016-01-25
RpcClient: Make reply-to queue name configurable

theomega:
Until now the queue name where the RpcClient expects the answers to be put is not configurable. The queue is statically set to `amq.rabbitmq.reply-to`.

This commit changes the two places where this queue name is used (for sending the request and when listening for responses): A member variable is used instead of the constant string. Default constructors are added so this **keeps backwards compatibility**.

The server doesn't need to be adapted: It takes the queue name from the header which the client puts in.

### DIFF
--- a/src/main/java/com/rabbitmq/client/RpcClient.java
+++ b/src/main/java/com/rabbitmq/client/RpcClient.java
@@ -48,6 +48,8 @@ public class RpcClient {
     private final String _exchange;
     /** Routing key to use for requests */
     private final String _routingKey;
+    /** Queue where the server should put the reply */
+    private final String _replyTo;
     /** timeout to use on call responses */
     private final int _timeout;
     /** NO_TIMEOUT value must match convention on {@link BlockingCell#uninterruptibleGet(int)} */
@@ -65,17 +67,20 @@ public class RpcClient {
      * Construct a new RpcClient that will communicate on the given channel, sending
      * requests to the given exchange with the given routing key.
      * <p/>
-     * Causes the creation of a temporary private autodelete queue.
+     * Causes the creation of a temporary private autodelete queue. The name of this queue can be specified.
      * @param channel the channel to use for communication
      * @param exchange the exchange to connect to
      * @param routingKey the routing key
+     * @param replyTo the queue where the server should put the reply
      * @param timeout milliseconds before timing out on wait for response
      * @throws IOException if an error is encountered
      */
-    public RpcClient(Channel channel, String exchange, String routingKey, int timeout) throws IOException {
+    public RpcClient(Channel channel, String exchange, String routingKey, String replyTo, int timeout) throws
+            IOException {
         _channel = channel;
         _exchange = exchange;
         _routingKey = routingKey;
+        _replyTo = replyTo;
         if (timeout < NO_TIMEOUT) throw new IllegalArgumentException("Timeout arguument must be NO_TIMEOUT(-1) or non-negative.");
         _timeout = timeout;
         _correlationId = 0;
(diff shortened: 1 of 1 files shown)


michaelklishin:
I assume you understand that the default uses [Direct Reply-to](https://www.rabbitmq.com/direct-reply-to.html), which is different to "client-driven" request-response? In other words, it's not just a default name, it is a special server-supported feature you are giving up if you use a different name.

theomega:
Yeah, I know.
The problem is that we have to support pre 3.4 servers, where the direct reply to feature is not available. I wanted to use an random queue in this case by declaring it and then passing in the queue name.

Or do you suggest something else? I could also declare the `amq.rabbitmq.reply-to` queue, but then all responses (on pre 3.4 servers) end in the same queue.

michaelklishin:
I assume some would want to use the most recent client version to get automatic recovery and improvements, so fair enough. As long as this is backwards compatible, we can merge it.

michaelklishin:
I've noticed that a comment mentions a queue named `"amq.rabbitmq.reply-to"`. That does not actually declare a queue, that's a special alias. I will correct it.

theomega:
How would using a "most recent client" help me? I'm actually on a recent client, but with an old server the queue is not there and I cannot declare it.

michaelklishin:
@theomega you can use an older client library version with older server releases. You can also use the most recent client release and the reason to do so is not the use of Direct Reply-to in this class. It is much more likely than those using recent client versions with old server versions are looking for automatic connection recovery.

If this wasn't the case, I'd simply tell you to use an older client version and close this.

src/main/java/com/rabbitmq/client/RpcClient.java
```diff
@@ -96,9 +118,27 @@ public RpcClient(Channel channel, String exchange, String routingKey, int timeou
...
+     * Causes the creation of a temporary private autodelete queue. The name of this queue will be
+     * "amq.rabbitmq.reply-to".
+     * @param channel the channel to use for communication
+     * @param exchange the exchange to connect to
+     * @param routingKey the routing key
+     * @param timeout milliseconds before timing out on wait for response
+     * @throws IOException if an error is encountered
+     */
+    public RpcClient(Channel channel, String exchange, String routingKey, int timeout) throws IOException {
+        this(channel, exchange, routingKey, "amq.rabbitmq.reply-to", timeout)
```
michaelklishin: This does not compile.

theomega:
@michaelklishin When using the most recent client with a pre 3.4 server, you get the folllowing exception:

    Caused by: com.rabbitmq.client.ShutdownSignalException: channel error; protocol method: #method<channel.close>(reply-code=404, reply-text=NOT_FOUND - no queue 'amq.rabbitmq.reply-to' in vhost 'prod', class-id=60, method-id=20)

You cannot fix this, because I cannot manually declare the `amq.rabbitmq.reply-to` queue because of the protected prefix `amq`. So there is no way of getting the current RpcClient to run with a pre 3.4 server. This is the reason for this commit.

Or did I get something wrong?

src/main/java/com/rabbitmq/client/RpcClient.java
theomega: Fixed, sorry!

michaelklishin:
@theomega OK, let me try again. As a maintainer, I have two options right now:

 * Tell you to use an older client
 * Merge this change

The former would prevent you from using automatic connection recovery, at least with all the fixes we did there ever since it was introduced. So the latter option is reasonable. Otherwise it wouldn't be.

Clearer now?

theomega:
Agree, using an old client actually never came to my mind, mainly because we have a mixed environment with newer servers.

Also agree that you as a maintainer need to decide to the best of the project, do it what you think is better.

Thanks for you help!

michaelklishin:
Merged into master manually with compilation and documentation corrections. Will be in nightly builds tomorrow and `3.7.0` (we could get it into `3.6.x` but this was submitted against `master`). Let me know if getting this into `3.6.1` matters to you at all.

michaelklishin:
and thank you.

theomega:
Thanks!

michaelklishin:
I cherry-picked this to `stable`, going to test a bit and hopefully push, so this will be in `3.6.1`.

PR #133, opened 2016-02-06
Added Closable-inteface to Connection-class for future try-with-resou…

### LINKED ISSUE #131: Implement Closable interface
Channel and Connection can implement the Closable interface, the close()-method is already implemented. This will open up the possibility to compile with Java 7 and use the try-with-resource function to avoid try-catch in finally-block of close

Khazrak:
Fixes #131.

Added Closable in the Connection-class. This will enable try-with-resources if somebody compiles it with JDK7 or later.

I wanted to add it to Channel-class also but it had a checked exception TimedOutException that broke the contract of Closeable so I avoided it.

### DIFF
--- a/src/main/java/com/rabbitmq/client/Connection.java
+++ b/src/main/java/com/rabbitmq/client/Connection.java
@@ -16,6 +16,7 @@
 
 package com.rabbitmq.client;
 
+import java.io.Closeable;
 import java.io.IOException;
 import java.net.InetAddress;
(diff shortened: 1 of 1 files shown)


michaelklishin:
@Khazrak thank you. Just like with #132, this changes an interface so as much as I'd like to get it into a future `3.6.x` release, it should probably ship with `3.7.0` only.

Issue #135, opened 2016-03-08
AutorecoveringConnection fires client shutdown hooks after connection has recovered


doswell:
Version - amqp-client-3.6.1

When init is called on AutorecoveringConnection it adds the automatic recovery listener to the shutdownHooks list - Because this is the first entry, any user added shutdown hooks will be called after the automatic recovery. 

as a possible solution;
The listener created in addAutomaticRecoverListener would likely need to call each shutdownListener, and omit adding the created listener to the shutdownHooks. 
addShutdownListener, removeShutdownListener would only apply to the shutdownHooks list and not add to the delegate. 
recoverShutdownListeners would just call addAutomaticRecoverListener.

michaelklishin:
I'd like to see if this is in line with the idea of #132 first.

michaelklishin:
Also note that shipping this in a `3.6.x` release is probably unrealistic.

doswell:
I think this fix allows for #132 to be easily done without other API
changes. As now you will get the shutdown listener fired when recovery
starts.
As a side note; Perhaps in terms of the auto recovery connection
shutdownlistener is a slightly misleading name vs disconnect listener.

Currently, if you add a shutdown listener which just logs, and are
connecting to one broker (I suspect if you use a cluster this is less
noticeable as the reconnect happens quickly to another member, so if
logging disconnects it seems correct);

Client Connect to server
Add shutdown listener and recovery listener to log status
(disconnect/connected)
Stop rabbitmq
-- no log output
Wait 5mins

Start rabbitmq
Client reconnects
-- logs output saying connected
-- logs output saying disconnected(for disconnect 5mins ago)

Funny enough the logging order was what I have seen from my tests; I think
the recovery listener ends up being fired before letting the remaining
shutdownlisteners run.

michaelklishin:
@doswell ok, thanks for the explanation, the latter point is definitely convincing.

michaelklishin:
@garyrussell @artembilan @spatula75 do you folks have an opinion on this change? It's breaking but makes sense to me. There is a pull request in #137.

artembilan:
@michaelklishin , 
thank you for letting to know!

I would like to see the direct code changes to make a decision. Right now it is difficult to dig over all those GH cross links.
But from other side I believe @garyrussell  in his comment: https://github.com/rabbitmq/rabbitmq-java-client/pull/132#issuecomment-180559626.
[...]

PR #136, opened 2016-03-08
Update AutorecoveringConnection to call ShutdownListeners before reco…

doswell:
…nnecting.

Re; #135

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
@@ -7,12 +7,15 @@ import com.rabbitmq.client.Channel;
 import com.rabbitmq.client.Connection;
 import com.rabbitmq.client.MissedHeartbeatException;
 import com.rabbitmq.client.Recoverable;
+import com.rabbitmq.client.RecoverableConnection;
+import com.rabbitmq.client.RecoverableShutdownSignalException;
 import com.rabbitmq.client.RecoveryListener;
 import com.rabbitmq.client.ShutdownListener;
 import com.rabbitmq.client.ShutdownSignalException;
 import com.rabbitmq.client.TopologyRecoveryException;
 import com.rabbitmq.client.impl.ConnectionParams;
 import com.rabbitmq.client.ExceptionHandler;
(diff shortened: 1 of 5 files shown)


src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -456,9 +455,8 @@ synchronized private void beginAutomaticRecovery() throws InterruptedException,
     }
 
     private void recoverShutdownListeners() {
-        for (ShutdownListener sh : this.shutdownHooks) {
-            this.delegate.addShutdownListener(sh);
-        }
+    	//Add the auto recovering listener to the new delegate.
+    	addAutomaticRecoveryListener();
```
michaelklishin: I find it non-obvious how the (user-provided) hooks are recovered.

michaelklishin:
I need to think if I'm 100% happy with this particular implementation of the idea in #135.

doswell:
@michaelklishin  I'd probably say "no not happy" I've worked on a more clear, and probably more importantly, more consistent implementation of calling the user hooks. 
In short there is a ForwardingRecoveryListener" class where user hooks are added into, and which gets fired prior to the recovery starting. It also applies the same sync blocks that the existing delegate listeners use. I just have to get it up to git. 
The hooks are also just maintained within AutorecoveringConnection and not added to the underlying delegate.
I'll try and get that up tomorrow

michaelklishin:
@doswell thank you. One idea I had is simply subclassing `ShutdownListener` (to produce a marker class) so that we can tell a recovery shutdown hooks from others so that they can be filtered out. How does that sound?

Alternatively we could use annotations but then I'm afraid this cannot be backported into `stable`.

doswell:
I didn't see a need for the subclassing in terms of the shutdown listeners just notifying that a disconnect occurs. ie, keeping separate from #132. But I think the idea of subclassing may be useful for #132 in terms of having a RecoveryStartedListener / RecoveryCompletedListener which can register on the RecoveryListeners without needing API changes. 

Close for #141

michaelklishin:
Thank you for your time. There are some class responsibility and cosmetic things in this PRs that I disagree with but I think I understand the intent in #135 well enough now. I will look into an alternative PR for #135.

PR #140, opened 2016-04-04
Set the nowait flag in queueBindNoWait()

MikePlayle:
This prevents us from being confused by a reply we didn't expect.

Also fix the associated javadoc comment.


michaelklishin:
Thank you for catching this.

Issue #99, opened 2015-09-24
User specified connection name


michaelklishin:
Related to https://github.com/rabbitmq/rabbitmq-server/issues/104 but can be useful (and shipped) independently: we should provide a way for connections to be named. The name would then be used in thread names.

michaelklishin:
Unfortunately there's no time to fit this into `3.6.0`.

michaelklishin:
Closing in favour of https://github.com/rabbitmq/rabbitmq-server/issues/104 (in this particular client; others don't have to follow in lock step).

PR #142, opened 2016-04-12
Support client specified connection name

hairyhum:
Related to https://github.com/rabbitmq/rabbitmq-server/issues/104 and #99 
Overloaded `newConnection` to be able to specify connection name.
Added `getConnectionName` to `AMQConnection` to retrieve client property value or null.

### DIFF
--- a/src/com/rabbitmq/client/Connection.java
+++ b/src/com/rabbitmq/client/Connection.java
@@ -94,6 +94,13 @@ public interface Connection extends ShutdownNotifier { // rename to AMQPConnecti
     Map<String, Object> getClientProperties();
 
     /**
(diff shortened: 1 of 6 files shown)


src/com/rabbitmq/client/Connection.java
```diff
@@ -94,6 +94,13 @@
     Map<String, Object> getClientProperties();
 
     /**
+     * Get connection name client property value
+     *
+     * @return string connection name from client properties, or null if there is not such property.
+     */
+    String getConnectionName();
```
michaelklishin: It should be `getClientProvidedConnectionName` or `getCustomConnectionName`. I will do the renaming.

PR #144, opened 2016-04-14
Idea: recovery started listener gets called when recovery begins

spatula75:
Recent rebase from master hopefully makes this cleaner.

This is still in a rough "idea" kind of state, just to communicate the notion of having an event to fire immediately prior to recovery beginning, in case someone wants to do something when that happens.  Might not be valuable anymore since ShutdownHooks are going to fire first (provided we can tell somehow that a recovery is going to happen).

### DIFF
--- a/src/main/java/com/rabbitmq/client/RecoveryListener.java
+++ b/src/main/java/com/rabbitmq/client/RecoveryListener.java
@@ -8,4 +8,5 @@ package com.rabbitmq.client;
  */
 public interface RecoveryListener {
     public void handleRecovery(Recoverable recoverable);
+    public void handleRecoveryStarted(Recoverable recoverable);
 }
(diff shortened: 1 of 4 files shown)


michaelklishin:
@spatula75 has this developed beyond a "rough idea"? Should we investigate it (maybe in a couple of weeks)?

spatula75:
Sadly I've been mired in other work and haven't had the time to direct the evolution of the idea or keep it up to date with the rest of the evolution of RabbitMQ.

michaelklishin:
@spatula75 that's perfectly fine. I guess my question really is, do you think this is worth pursuing? It is something you've been using in production or feel the need for strongly?

spatula75:
I think it would still be very helpful in our code to know if a recovery is currently in progress (or has started), by whichever means is the most intuitive and clean with respect to the rest of the client code.

vikinghawk:
+1 to this as well. Getting the recovery notification before the error notification seems out of order and is confusing to our support people when they see the recovery notification event in the logs before the actual error that caused connection loss.

I have a similar use case where we want to know as soon as possible that a connection was lost and recovery is beginning. I was able to sort of make this work by implementing my own ExceptionHandler and keying off a handleConnectionRecoveryException() notification to get a good idea that a recovery is in progress. However this only works for usecases where the first reconnect attempt fails. If it is able to reconnect on its first attempt the handleConnectionRecoveryException method is never called and I just get a RevoveryListener event and then ShutdownListener event.

acogoluegnes:
Merged manually. Thanks @spatula75!

spatula75:
Not sure if I should say "you're welcome" or be terrified!  :-D

michaelklishin:
FWIW I like this idea :)

PR #164, opened 2016-08-09
Finish transition to Maven

dumbbell:
We already use Maven to compile the library, but several parts were still based on Ant. This pull request finishes the transition to fully use Maven.


michaelklishin:
👍

Issue #162, opened 2016-08-02
JUnit gets dragged into OSGi manifest


cfiehe:
Since release 3.6.4 of amqp-client the following two packages get imported in the manifest header:
- junit.framework
- org.junit

The bundle cannot be resolved in an OSGi framework, because the test packages are not provided in a productive OSGi environment.

Version 3.6.3 of amqp-client does not import these two packages and gets resolved without any issues.

OSGi compatibility is a very important feature for me. Can someone please have a look what's going on during the build process.

cfiehe:
I have posted a small Maven-based project in the RabbitMQ users group. This is the link:
https://groups.google.com/d/msg/rabbitmq-users/MsHYLoa9qEo/Sqfd6LlJBwAJ

Hopefully, it helps you to fix the problem.

michaelklishin:
Note: to get into `3.6.x`, this depends on our build infra being fully migrated to Maven. However, it is fixed in master by @cfiehe => closing.

PR #165, opened 2016-08-10
Restoring OSGi compatibility after migration to a fully Maven-based b…

cfiehe:
…uild process. This commit also fixes issue #162.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -9,7 +9,7 @@
   <packaging>jar</packaging>
 
   <name>RabbitMQ Java Client</name>
-  <description>RabbitMQ Java client</description>
+  <description>The RabbitMQ Java client library allows Java code to interface to RabbitMQ.</description>
   <url>http://www.rabbitmq.com</url>
 
   <licenses>
@@ -43,6 +43,11 @@
     <url>https://github.com/rabbitmq/rabbitmq-java-client.git</url>
     <connection>scm:git:https://github.com/rabbitmq/rabbitmq-java-client.git</connection>
   </scm>
+  
+  <organization>
+    <name>Pivotal Software, Inc.</name>
+    <url>http://www.rabbitmq.com</url>
+  </organization>
 
   <properties>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -9,7 +9,7 @@
   <packaging>jar</packaging>
 
   <name>RabbitMQ Java Client</name>
-  <description>RabbitMQ Java client</description>
+  <description>The RabbitMQ Java client library allows Java code to interface to AMQP servers.</description>
```
michaelklishin: Can we please back this change out?

pom.xml
```diff
@@ -43,6 +43,11 @@
     <url>https://github.com/rabbitmq/rabbitmq-java-client.git</url>
     <connection>scm:git:https://github.com/rabbitmq/rabbitmq-java-client.git</connection>
   </scm>
+  
+  <organization>
+    <name>Rabbit Technologies Ltd.</name>
```
michaelklishin: Rabbit Technologies has been acquired by SpringSource/VMware in 2010. Please use `Pivotal Software, Inc`.

pom.xml
michaelklishin: Or at least replace "AMQP servers" with "RabbitMQ" because "AMQP" means more than one protocol these days.

michaelklishin:
Thank you!

pom.xml
cfiehe: I have just copied the description from here: https://www.rabbitmq.com/java-client.html. But no problem, I will replace "AMQP servers" with "RabbitMQ". The description in the pom gets used as "Bundle-Description" in the OSGi manifest file and things look better with a proper outline.

pom.xml
cfiehe: I was not aware of that fact. It will be changed immediately.

cfiehe:
The pull request has been updated.

michaelklishin:
Thank you once again. We will be backporting all the Maven bits to `stable` and this client releases will be decoupled from RabbitMQ releases fairly soon. No promises on whether this will make it into `3.6.6` but chances are it will.

cfiehe:
That was pretty fast ;-) Thank you very much for the merge.

No problem.., From my side, I can stay on version 3.6.3 until the fix gets it into the release cycle.

PR #170, opened 2016-08-23
ID for PerfTest jobs and Unique system out for better output in multithreading.

abousselmi:
Dear reviewer,
I added an ID variable to PerfTest.java to identify each running test.
I also modified the System.out to work better when multiple PerfTest threads are running in the same machine.
Best regards,
AB.


michaelklishin:
I assume this can be released in `3.6.x`? If so, please also submit this against `stable`.

michaelklishin:
Thank you. I am not 100% sure we should ship this in `3.6.x` so it will have to wait for 4.0.0 (née 3.7.0).

abousselmi:
Thank you for your reply. Until then!

acogoluegnes:
@abousselmi maybe you could rename `showRate` to `getRate` now it returns a String and no longer show something in the console?

abousselmi:
It's a good idea! Will do it. Thank you.

abousselmi:
@acogoluegnes I added a test id also to the final output messages, whould it be possible to make a new pull request for it ? Thank you.

michaelklishin:
@abousselmi you can submit a new pull request, sure.

abousselmi:
Thank you for you reply.

PR #173, opened 2016-08-24
Create enum for exchange types

### LINKED ISSUE #150: Create enum or constants for exchange types.
I would like to see an enum or constants within a class for the exchange types which (as I know) are `amq.direct`, `amq.fanout`, `amq.topic`, `amq.match` and `amq.headers`.

acogoluegnes:
Fixes #150

### DIFF
--- a/src/main/java/com/rabbitmq/client/ExchangeType.java
+++ b/src/main/java/com/rabbitmq/client/ExchangeType.java
@@ -1,0 +1,19 @@
+package com.rabbitmq.client;
+
+/**
+ *
+ */
+public enum ExchangeType {
+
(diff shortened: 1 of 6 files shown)


src/main/java/com/rabbitmq/client/ExchangeType.java
```diff
@@ -0,0 +1,19 @@
+package com.rabbitmq.client;
+
+/**
+ *
+ */
+public enum ExchangeType {
```
michaelklishin: I suggest that we rename this to `BuiltinExchangeType` because exchange types are certainly not limited to these 4.

michaelklishin:
We will merge this and cherry-pick to `stable`.

PR #175, opened 2016-08-24
Add AddressResolver to try all DNS record IPs

### LINKED ISSUE #138: Make socket connection to try all DNS record IPs
In a HA RabbitMQ cluster, I use DNS round robin technic to load balance the load between nodes and replace HAProxy, for instance:

rabbitmq => 10.0.0.1 10.0.0.2 10.0.0.3

It works like a charm with the Php extension, because here:

https://github.com/alanxz/rabbitmq-c/blob/master/librabbitmq/amqp_socket.c#L449

We can see that all IP adresses from DNS record are "tested", so if a node goes down, no problemo, no error on Php side when connecting. 

But, I also use logstash to consume RabbitMQ messages, which uses this library, and apparently, this little trick is not working as I can see on:

https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/FrameHandlerFactory.java#L28

Do you think this could be a feature here (or I may miss a Java trick here) ?

Thanks you,

acogoluegnes:
Fixes #138

### DIFF
--- a/src/main/java/com/rabbitmq/client/AddressResolver.java
+++ b/src/main/java/com/rabbitmq/client/AddressResolver.java
@@ -1,5 +1,6 @@
 package com.rabbitmq.client;
 
+import java.io.IOException;
 import java.util.List;
 
 /**
@@ -7,6 +8,6 @@ import java.util.List;
  */
 public interface AddressResolver {
 
-    List<Address> getAddresses();
+    List<Address> getAddresses() throws IOException;
 
 }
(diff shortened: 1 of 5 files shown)


michaelklishin:
Should this be used by `ConnectionFactory` by default?

acogoluegnes:
@ebuildy @michaelklishin here's a first cut of the feature. It's not used by default right now, I was thinking the following:

* using the `DnsRecordIpAddressResolver` by default in master (4.0.0) for the non-array `ConnectionFactory.newConnection` methods
* providing the `DnsRecordIpAddressResolver` in stable (3.6.x), but not using it by default

 WDYT?

michaelklishin:
@acogoluegnes sounds reasonable to me.

PR #180, opened 2016-08-26
Add SLF4J

### LINKED ISSUE #114: Logging library integration
Over the years the library has been 0-dependency but issues such as #113 and calls to introduce Metrics integration are pretty common now.

This issue is only about logging. Should we use SLF4J? Or some kind of handlers for apps to hook into (was discussed for .NET client, for example). It remains to be discussed and decided.

acogoluegnes:
Fixes #114

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -53,6 +53,10 @@
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
+    <slf4j.version>1.7.21</slf4j.version>
+    <commons-cli.version>1.1</commons-cli.version>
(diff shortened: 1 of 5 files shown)


acogoluegnes:
We now have a bunch a stack traces bubbling up during the test suite, because some errors were not logged at all in the exception handler. They should be harmless, as they occur during tests of error scenario. Let's see after a few builds if we should remove them, to make the CI reports easier to read.

michaelklishin:
@acogoluegnes this can be handled separately but I think `java.net.SocketException`s with reason "Socket closed" and even "Connection reset" should be logged as warnings: they are just a fact of life in any production system. Logging them as errors will scare too many users for no good reason.

So logging them as warnings and maybe without full stack trace would be sufficient. Feel free to look into it in a follow-up PR.

PR #181, opened 2016-08-26
Refactor tests in TestMain

### LINKED ISSUE #108: Port relevant TestMain parts to JUnit
`com.rabbitmq.examples.TestMain` is used as an integration test that doesn't use JUnit and tests things many of which are already covered by the rest of the suite. We need to move relevant bits to JUnit and remove the rest.

acogoluegnes:
Fixes #108


src/test/java/com/rabbitmq/client/test/functional/FrameMax.java
```diff
@@ -50,6 +51,10 @@ public FrameMax() {
         connectionFactory.setRequestedFrameMax(FRAME_MAX);
     }
 
+    @Test public void negociationOk() {
```
michaelklishin: I believe this is meant to say "negotiation"?

acogoluegnes: Yes, sorry, French spelling.

michaelklishin:
@acogoluegnes I believe this can be backported to `stable`? If so, please do it.

PR #186, opened 2016-08-29
Introduce metrics support

### LINKED ISSUE #79: Add JMX monitoring
It would be very useful to get at least some basic statistics published via JMX. E.g. number of messages published, consumed, rejected, number of threads used etc.

acogoluegnes:
Fixes #79


src/main/java/com/rabbitmq/client/impl/MetricsStatistics.java
```diff
@@ -21,128 +53,306 @@
...
-    public long getConnectionCount() {
-        return connections.getCount();
+    public void newConnection(final Connection connection) {
+        try {
+            if(connection.getId() == null) {
+                connection.setId(UUID.randomUUID().toString());
+            }
+            connections.inc();
+            connectionState.put(connection.getId(), new ConnectionState(connection));
+            connection.addShutdownListener(new ShutdownListener() {
```
michaelklishin: We need to make sure that we don't accumulate more than one listener between connection recoveries.

acogoluegnes: `ShutdownListener` is not added to the `AutorecoveringConnection` now, only to actual connections, see 321c3bd. Found a bug BTW, thanks.

src/test/java/com/rabbitmq/client/test/functional/Statistics.java
```diff
@@ -0,0 +1,552 @@
...
+            connections[i] = connectionFactory.newConnection();
+            for(int j=0;j<nbChannelsPerConnection;j++) {
+                Channel channel = connections[i].createChannel();
+                channel.basicQos(1);
+                channels[i*nbChannelsPerConnection+j] = channel;
+            }
+        }
+
+        // consume messages without ack
+        for(int i=0;i<nbOfMessages;i++) {
```
michaelklishin: Please use spaces: `for(int i = 0; i < nbOfMessages; i++)`.

src/test/java/com/rabbitmq/client/test/functional/Statistics.java
```diff
@@ -0,0 +1,552 @@
...
+        int nbTasks = nbChannels; // channel are not thread-safe
+
+        Random random = new Random();
+
+        // create connections
+        Connection [] connections = new Connection[nbConnections];
+        Channel [] channels = new Channel[nbChannels];
+        for(int i=0;i<nbConnections;i++) {
+            connections[i] = connectionFactory.newConnection();
+            for(int j=0;j<nbChannelsPerConnection;j++) {
```
michaelklishin: Same as below.

src/main/java/com/rabbitmq/client/StatisticsCollector.java
```diff
@@ -0,0 +1,52 @@
...
+/**
+ * Interface to gather execution data of the client.
+ * Note transactions are not supported: they deal with
+ * publishing and acknowledgments and the collector contract
+ * assumes then that published messages and acks sent
+ * in a transaction are always counted, even if the
+ * transaction is rolled back.
+ *
+ */
+public interface StatisticsCollector {
```
michaelklishin: I suggest that we rename it to `MetricsCollector` and the Metrics library-based implementation then becomes `StandardMetricsCollector`.

Issue #11, opened 2013-10-08
Switch to NIO[2]


michaelklishin:
Reported and discussed in some depth on [rabbitmq-discuss](http://rabbitmq.1065348.n5.nabble.com/Long-timeout-if-server-host-becomes-unreachable-td30275.html).

Internal bug tracker: `#23327`.

The only solution is switching to NIO. NIO would yield little improvement for the client the way it works today but worth moving to for this issue.

arunvisw:
did we get this feature in?

michaelklishin:
@arunvisw we close issues that are fixed or no longer relevant for other reasons. This one is not closed ;)

arunvisw:
Would really appreciate if you can suggest some work arounds

michaelklishin:
The [original thread](http://rabbitmq.1065348.n5.nabble.com/Long-timeout-if-server-host-becomes-unreachable-td30275.html) mentions a workaround.

Hronom:
I were any progress on solving blocks by timeout?

michaelklishin:
@Hronom I don't understand what "blocks by timeout" means but there is no way to avoid `Socket#write` blocking in some cases without a really ugly workaround provided in a thread linked above. The right thing to do is to switch to NIO and since this issue is open, that is not done.

Hronom:
So switching to NIO helps to solve situation when `Channel#basicPublish` block thread where it called without any notification about that connection slowed down by flow state?

michaelklishin:
@Hronom please get familiar with the original mailing list report. This has no relation to internal flow control (or generally speaking, flow control period). While flow control may make this more visible, it can be reproduced without flow control.

michaelklishin:
We believe the original issue is addressed in https://github.com/rabbitmq/rabbitmq-java-client/pull/191 but we'll keep this issue open because switching to NIO is still something we are interested in.

PR #191, opened 2016-09-01
Prevent deadlock/hanging on JDK socket write

### LINKED ISSUE #194: Prevent Socket.write() from blocking
Originally reported on [rabbitmq-discuss](http://rabbitmq.1065348.n5.nabble.com/Long-timeout-if-server-host-becomes-unreachable-td30275.html)

Java NIO could be a solution to blocking on writes, but it's a major rewriting of the client. This migration is considered on the mid-term, as it could bring features and reliability to the table.

In the meantime, a lighter solution to the problem mentioned above could be found, e.g. making part of the connection shutdown (where the blocking/deadlock occur) asynchronous.

acogoluegnes:
Add a timeout do some shutdown operations as they can deadlock (mainly because JDK sockets don't support a timeout on write operations). This can solve the problem which gave birth to #11, and thus avoids a complete NIO2 rewrite.

Fixes #194 
References #11

### DIFF
--- a/src/main/java/com/rabbitmq/client/ConnectionFactory.java
+++ b/src/main/java/com/rabbitmq/client/ConnectionFactory.java
@@ -642,7 +642,7 @@ public class ConnectionFactory implements Cloneable {
     }
 
     protected FrameHandlerFactory createFrameHandlerFactory() throws IOException {
-        return new FrameHandlerFactory(connectionTimeout, factory, socketConf, isSSL());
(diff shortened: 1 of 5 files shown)


michaelklishin:
@acogoluegnes thanks. We can do this for 4.0 but we really should switch to NIO in the medium term in my opinion.

michaelklishin:
@acogoluegnes it looks reasonable at first glance.

PR #192, opened 2016-09-02
Log UnknownChannelException in connection loop

### LINKED ISSUE #189: UnknownChannelException in AMQConnection.MainLoop.run
Encountered this error in my ExceptionHandler.handleUnexpectedConnectionDriverException method a couple times on version 3.6.3 with connection and topology recovery enabled.

Message: RabbitExceptionHandler: Caught unexpected connection driver exception for connection XXX
Throwable: com.rabbitmq.client.impl.UnknownChannelException: Unknown channel number 147
	at com.rabbitmq.client.impl.ChannelManager.getChannel(ChannelManager.java:82)
	at com.rabbitmq.client.impl.AMQConnection$MainLoop.run(AMQConnection.java:560)

Looks to me like there is possibility MainLoop is processing a frame with a channel number that doesn't exist in ChannelManager? Not sure how that is possible and I can't reliably reproduce... Could potentially handle this scenario in a better way though?

acogoluegnes:
Fixes #189

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/AMQConnection.java
+++ b/src/main/java/com/rabbitmq/client/impl/AMQConnection.java
@@ -15,14 +15,6 @@
 
 package com.rabbitmq.client.impl;
 
-import java.io.EOFException;
-import java.io.IOException;
-import java.net.InetAddress;
-import java.net.SocketException;
-import java.net.SocketTimeoutException;
-import java.util.*;
-import java.util.concurrent.*;
-
 import com.rabbitmq.client.*;
 import com.rabbitmq.client.Method;
 import com.rabbitmq.client.impl.AMQChannel.BlockingRpcContinuation;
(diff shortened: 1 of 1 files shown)


src/main/java/com/rabbitmq/client/impl/AMQConnection.java
```diff
@@ -554,7 +559,17 @@ public void run() {
...
                                     if (cm != null) {
-                                        cm.getChannel(frame.channel).handleFrame(frame);
+                                        ChannelN channel;
+                                        try {
+                                            channel = cm.getChannel(frame.channel);
+                                        } catch(UnknownChannelException e) {
+                                            // this can happen if channel has been closed,
+                                            // but there was e.g. an in-flight delivery.
+                                            // just ignoring the frame to avoid closing the whole connection
+                                            LOGGER.info("Channel {} is unknown, ignoring frame", frame.channel);
```
michaelklishin: I'd change the wording to "Received a frame on an unknown channel, ignoring it"

src/main/java/com/rabbitmq/client/impl/AMQConnection.java
```diff
@@ -554,7 +559,17 @@ public void run() {
...
                                     if (cm != null) {
-                                        cm.getChannel(frame.channel).handleFrame(frame);
+                                        ChannelN channel;
+                                        try {
+                                            channel = cm.getChannel(frame.channel);
+                                        } catch(UnknownChannelException e) {
+                                            // this can happen if channel has been closed,
+                                            // but there was e.g. an in-flight delivery.
+                                            // just ignoring the frame to avoid closing the whole connection
+                                            LOGGER.info("Received a frame on an unknown channel, ignoring it");
```
michaelklishin: Sorry, I didn't mean that you should use this message exactly. I think logging channel ID of the frame is still a good idea :)

michaelklishin:
Will update the copy myself.

PR #197, opened 2016-09-15
Recovery related changes

vikinghawk:
This PR contains some changes related to recovery. Note i cleaned up a bunch of compiler warnings by adding the missing Override tags on methods. I can submit another PR without those changes if you guys would prefer.

Non-cosmetic changes are in AutorecoveringConnection and AutorecoveringChannel.

Fixes:
1) fixes a NPE that can occur when a thread attempts to create a new channel after a connection has been recovered but before channel recovery has completed. More details of issue at https://groups.google.com/forum/#!topic/rabbitmq-users/7P5fYQLLP6g. Fixes the NPE in AutorecoveringChannel, but then also changed the AutorecoveryingConnection to not set the connection delegate until channel recovery is complete. This allows channel recovery to complete without error at the expense of the separate thread attempting to create a new channel of getting an AlreadyClosedException because it would happen on the old delegate.

Other changes:
1) Added a check in recoverQueues() that skips over all the queue name change logic if the name remained the same.

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
@@ -103,17 +103,13 @@ public class AutorecoveringConnection implements Connection, Recoverable, Networ
      */
     public void init() throws IOException, TimeoutException {
         this.delegate = this.cf.newConnection();
-        this.addAutomaticRecoveryListener();
-    }
-
-    public void start() throws IOException {
-        // no-op, AMQConnection#start is executed in ConnectionFactory#newConnection
-        // and invoking it again will result in a framing error. MK.
+        this.addAutomaticRecoveryListener(delegate);
     }
 
     /**
      * @see com.rabbitmq.client.Connection#createChannel()
      */
+    @Override
     public Channel createChannel() throws IOException {
         RecoveryAwareChannelN ch = (RecoveryAwareChannelN) delegate.createChannel();
         if (ch == null) {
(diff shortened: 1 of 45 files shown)


michaelklishin:
Thank you!

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -103,17 +103,13 @@ public AutorecoveringConnection(ConnectionParams params, FrameHandlerFactory f,
      */
     public void init() throws IOException, TimeoutException {
         this.delegate = this.cf.newConnection();
-        this.addAutomaticRecoveryListener();
-    }
-
-    public void start() throws IOException {
```
michaelklishin: This method existed for a reason. Are you 100% sure it is safe to remove it now, or a good idea to do so?

michaelklishin: Digging in code history suggests that the method was introduced around the time we wanted to move Langohr and March Hare to use recovery implemented by this client, and that's why https://github.com/rabbitmq/rabbitmq-java-client/commit/1e603d9de9d2fa18689a305546bcefcf4d3e49c7 calls out Langohr specifically.

So far it looks like it's safe to delete this method.

michaelklishin: OK, Langohr recovery test suite works just fine with this change.

acogoluegnes:
@vikinghawk Thanks!

michaelklishin:
This was back ported for `3.6.6` => changing milestone.

PR #199, opened 2016-09-18
Fix for QUEUE_NOT_FOUND exceptions during recovery.

vikinghawk:
This is a follow up to [PR 197](https://github.com/rabbitmq/rabbitmq-java-client/pull/197) and is the fix for issue 1 mentioned at https://groups.google.com/forum/#!topic/rabbitmq-users/7P5fYQLLP6g.

When a consumer is canceled the code was removing its auto-delete queue from RecordedQueues list, but it wasn't removing that queue's bindings. When recovery ran it tried recovering bindings for queues that did not exist and caused recovery failures. Line 788 of AutorecoveringConnection is the fix.

I also fixed several potential threading related issues with collections where it was possible to get ConcurrentModificationExceptions.

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/TruncatedInputStream.java
+++ b/src/main/java/com/rabbitmq/client/impl/TruncatedInputStream.java
@@ -42,7 +42,7 @@ public class TruncatedInputStream extends FilterInputStream {
     }
 
     @Override
-    public void mark(int readlimit) {
+    public synchronized void mark(int readlimit) {
         super.mark(readlimit);
         mark = counter;
     }
@@ -71,7 +71,7 @@ public class TruncatedInputStream extends FilterInputStream {
     }
 
     @Override
-    public void reset() throws IOException {
+    public synchronized void reset() throws IOException {
         super.reset();
         counter = mark;
     }
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
@@ -16,11 +16,12 @@
 package com.rabbitmq.client.impl.recovery;
 
 import com.rabbitmq.client.*;
+import com.rabbitmq.utility.Utility;
 
 import java.io.IOException;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.CopyOnWriteArrayList;
 import java.util.concurrent.TimeoutException;
 
 /**
@@ -30,13 +31,13 @@ import java.util.concurrent.TimeoutException;
  * @since 3.3.0
  */
 public class AutorecoveringChannel implements Channel, Recoverable {
-    private RecoveryAwareChannelN delegate;
-    private AutorecoveringConnection connection;
-    private final List<ShutdownListener> shutdownHooks  = new ArrayList<ShutdownListener>();
-    private final List<RecoveryListener> recoveryListeners = new ArrayList<RecoveryListener>();
-    private final List<ReturnListener> returnListeners = new ArrayList<ReturnListener>();
-    private final List<ConfirmListener> confirmListeners = new ArrayList<ConfirmListener>();
-    private final List<FlowListener> flowListeners = new ArrayList<FlowListener>();
+    private volatile RecoveryAwareChannelN delegate;
+    private volatile AutorecoveringConnection connection;
+    private final List<ShutdownListener> shutdownHooks  = new CopyOnWriteArrayList<ShutdownListener>();
+    private final List<RecoveryListener> recoveryListeners = new CopyOnWriteArrayList<RecoveryListener>();
+    private final List<ReturnListener> returnListeners = new CopyOnWriteArrayList<ReturnListener>();
+    private final List<ConfirmListener> confirmListeners = new CopyOnWriteArrayList<ConfirmListener>();
+    private final List<FlowListener> flowListeners = new CopyOnWriteArrayList<FlowListener>();
     private int prefetchCountConsumer;
     private int prefetchCountGlobal;
     private boolean usesPublisherConfirms;
@@ -648,7 +649,7 @@ public class AutorecoveringChannel implements Channel, Recoverable {
     }
 
     private void notifyRecoveryListeners() {
-        for (RecoveryListener f : this.recoveryListeners) {
+        for (RecoveryListener f : Utility.copy(this.recoveryListeners)) {
             f.handleRecovery(this);
         }
     }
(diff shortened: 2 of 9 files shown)


src/main/java/com/rabbitmq/client/impl/TruncatedInputStream.java
```diff
@@ -42,7 +42,7 @@ public int available() throws IOException {
     }
 
     @Override
-    public void mark(int readlimit) {
+    public synchronized void mark(int readlimit) {
```
vikinghawk: the super mark/reset methods are synchronized, this is addressing a findbug warning

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
```diff
@@ -30,13 +30,13 @@
...
-    private RecoveryAwareChannelN delegate;
-    private AutorecoveringConnection connection;
-    private final List<ShutdownListener> shutdownHooks  = new ArrayList<ShutdownListener>();
-    private final List<RecoveryListener> recoveryListeners = new ArrayList<RecoveryListener>();
-    private final List<ReturnListener> returnListeners = new ArrayList<ReturnListener>();
-    private final List<ConfirmListener> confirmListeners = new ArrayList<ConfirmListener>();
-    private final List<FlowListener> flowListeners = new ArrayList<FlowListener>();
+    private volatile RecoveryAwareChannelN delegate;
+    private volatile AutorecoveringConnection connection;
+    private final List<ShutdownListener> shutdownHooks  = new CopyOnWriteArrayList<ShutdownListener>();
```
vikinghawk: ChannelN was already using CopyOnWriteArrayList for these so followed same pattern here.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -60,19 +62,19 @@
...
 
-    private final List<ShutdownListener> shutdownHooks  = new ArrayList<ShutdownListener>();
-    private final List<RecoveryListener> recoveryListeners = new ArrayList<RecoveryListener>();
-    private final List<BlockedListener> blockedListeners = new ArrayList<BlockedListener>();
+    private final List<ShutdownListener> shutdownHooks  = Collections.synchronizedList(new ArrayList<ShutdownListener>());
+    private final List<RecoveryListener> recoveryListeners = Collections.synchronizedList(new ArrayList<RecoveryListener>());
+    private final List<BlockedListener> blockedListeners = Collections.synchronizedList(new ArrayList<BlockedListener>());
 
     // Records topology changes
-    private final Map<String, RecordedQueue> recordedQueues = new ConcurrentHashMap<String, RecordedQueue>();
```
vikinghawk: ConcurrentHashMap is perfectly thread-safe, but there were a couple places below where the code was synchronizing on the concurenthashmap to prevent access to it. That won't work unless the entire map itself is synchronized so switching to synchronized hash maps here. A copy is created of the map anytime it is iterated over now to make iteration thread safe.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -672,22 +672,22 @@ private void recoverConsumers() {
...
     private void propagateQueueNameChangeToConsumers(String oldName, String newName) {
-        for (RecordedConsumer c : this.consumers.values()) {
+        for (RecordedConsumer c : Utility.copy(this.consumers).values()) {
             if (c.getQueue().equals(oldName)) {
                 c.setQueue(newName);
             }
         }
     }
 
-    synchronized void recordQueueBinding(AutorecoveringChannel ch,
```
vikinghawk: synchronized was on these next 4 methods in what appears to be an effort to protect recordedBindings. This is now handled by wrapping it in Collections.synchronized so removing the class level synchronizations here.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -784,7 +784,9 @@ void maybeDeleteRecordedAutoDeleteQueue(String queue) {
                     RecordedQueue q = this.recordedQueues.get(queue);
                     // last consumer on this connection is gone, remove recorded queue
                     // if it is auto-deleted. See bug 26364.
-                    if((q != null) && q.isAutoDelete()) { this.recordedQueues.remove(queue); }
+                    if((q != null) && q.isAutoDelete()) {
+                        deleteRecordedQueue(queue);
```
vikinghawk: here is the fix. This method already handles the logic of removing the recorded queue and it's bindings.

src/main/java/com/rabbitmq/client/impl/recovery/RecordedConsumer.java
```diff
@@ -57,7 +57,7 @@ public RecordedConsumer autoAck(boolean value) {
     }
 
     public String recover() throws IOException {
-        this.consumerTag = this.channel.basicConsume(this.queue, this.autoAck, this.consumerTag, false, this.exclusive, this.arguments, this.consumer);
+        this.consumerTag = this.channel.getDelegate().basicConsume(this.queue, this.autoAck, this.consumerTag, false, this.exclusive, this.arguments, this.consumer);
```
vikinghawk: all the recorded entities besides these 2 use getDelegate(). I think these should 2 as well otherwise it ends up duplicating the RecorcedConsumer/Queue and readding it to the recorded entity collections.

src/main/java/com/rabbitmq/utility/Utility.java
```diff
@@ -85,4 +90,32 @@ public static String makeStackTrace(Throwable throwable) {
...
+    
+    /**
+     * Synchronizes on the list and then returns a copy of the list that is safe to iterate over. Useful when wanting to do thread-safe iteration over
+     * a List wrapped in {@link Collections#synchronizedList(List)}.
+     *
+     * @param list
+     *            The list, which may not be {@code null}
+     * @return ArrayList copy of the list
+     */
+    public static <E> List<E> copy(final List<E> list) {
```
vikinghawk: these seemed like the appropriate place...

acogoluegnes:
@vikinghawk Thanks!

michaelklishin:
Should this be backported to `3.6.x`?

acogoluegnes:
@michaelklishin Yes, we can. https://github.com/rabbitmq/rabbitmq-java-client/pull/197 should also be backported then.

vikinghawk:
Not sure on what the release dates for 3.6.x vs 4.0.0 are, but the sooner my team could get these changes in a released form the better. Thanks!

michaelklishin:
We will backport.

acogoluegnes:
It's been backported to stable, so it will be in 3.6.6.

michaelklishin:
This was back ported to `3.6.6` => changing milestone.

PR #209, opened 2016-10-21
Clean AutorecoveringConnection consumers correctly

### LINKED ISSUE #208: When a channel is closed, consumers should be cleaned up from AutorecoveringConnection cache
This can lead to memory leaks if many consumers are created and the consumers are not cancelled. If the channel is closed, the recovery would fail anyway.

See https://groups.google.com/forum/#!topic/rabbitmq-users/d9PKyK9WXsw

acogoluegnes:
Fixes #208

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
@@ -18,8 +18,7 @@ package com.rabbitmq.client.impl.recovery;
 import com.rabbitmq.client.*;
 
 import java.io.IOException;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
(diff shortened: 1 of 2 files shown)


michaelklishin:
@acogoluegnes please merge this into master (there are conflicts).

vikinghawk:
When a channel is recovered it gets a new consumerTag right?  I'm thinking AutorecoveringConnection.recoverConsumers() should remove the old tag and add the new one to the AutorecoveringChannel.consumerTags set.

acogoluegnes:
Good catch! Should be done around here: https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java#L659

I'll work on that, thanks for reporting!

Issue #217, opened 2016-11-23
Consider Adding toString() to AutorecoveringChannel


garyrussell:
Logging `Recoverable` in the `RecoveryListener` emits...

    Channel recovery complete: com.rabbitmq.client.impl.recovery.AutorecoveringChannel@3571b748

Consider adding something like:

    @Override
    public String toString() {
        return "AutorecoveringChannel:" + delegate;
    }

Workaround:

    ((AutorecoveringChannel) recoverable).getDelegate().toString();

michaelklishin:
Makes sense. Would you like to submit a PR? :)

garyrussell:
Will do

PR #218, opened 2016-11-23
GH-217: Add toString() to AutorecoveringChannel

garyrussell:
Resolves https://github.com/rabbitmq/rabbitmq-java-client/issues/217

I didn't add a prefix; to be consistent with `AutrorecoveringConnection`.


michaelklishin:
Thank you!

PR #220, opened 2016-12-01
Add configurable timeout on RPC calls

### LINKED ISSUE #219: Make continuation timeouts configurable (e.g. per channel)
See [this rabbitmq-users thread](https://groups.google.com/forum/#!topic/rabbitmq-users/Xc1TH-518qY).

There's already a way to execute continuations with a timeout but `Channel` doesn't expose it to the user. Note that we don't want to have N methods that duplicate the existing ones but also accept a timeout: it should be a one time setting.

acogoluegnes:
Fixes #219


src/main/java/com/rabbitmq/client/ChannelRpcTimeoutException.java
```diff
@@ -0,0 +1,35 @@
+package com.rabbitmq.client;
+
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * Exception thrown when a channel times out on a RPC call.
+ * @since 4.1.0
+ */
+public class ChannelRpcTimeoutException extends IOException {
```
michaelklishin: Please rename this to `ChannelContinuation…`. `RPC` will mislead those who use higher level libraries that may claim to support "RPC".

src/main/java/com/rabbitmq/client/ConnectionFactory.java
```diff
@@ -73,6 +73,9 @@
      *  zero means wait indefinitely */
     public static final int    DEFAULT_SHUTDOWN_TIMEOUT = 10000;
 
+    /** The default timeout for RPC calls in channels: no timeout */
+    public static final int    DEFAULT_CHANNEL_RPC_TIMEOUT = -1;
```
michaelklishin: How about we make it, say, 10 minutes instead of infinity?

src/main/java/com/rabbitmq/client/ChannelContinuationTimeoutException.java
```diff
@@ -0,0 +1,35 @@
...
+/**
+ * Exception thrown when a channel times out on a continuation during a RPC call.
+ * @since 4.1.0
+ */
+public class ChannelContinuationTimeoutException extends IOException {
+
+    /**
+     * The channel that performed the call.
+     */
+    private final Object channel;
```
michaelklishin: Should this be `Channel` or am I missing something?

acogoluegnes: This exception is thrown from `AMQChannel`, which is not a `Channel`. There are several solutions: 
 * having a `Channel` property in the exception and casting the `AMQChannel` to `Channel` when throwing the exception. We risk in theory a `ClassCastException`.
 * having an `AMQChannel` property in the exception, but the implementation leaks up to error handling.
 * having just the channel number and the connection in the exception, which should be enough to diagnose any error.

I think the third solution is a good compromise.

michaelklishin: Let's add a channel number field (we can keep the field we have with a comment).

Issue #52, opened 2015-05-04
Please handle InterruptedExceptions correctly


vfelberg:
Class com.rabbitmq.utility.BlockingCell has several while loops catching InterruptedExceptions and ignoring them. This makes it impossible to shutdown the app correctly; you have to kill it brutally. The correct handling of an InterruptedException in a while loop would be:

``` java
try {
  ...
} catch (InterruptedException e) {
  Thread.currentThread().interrupt();
  return;
}
```

There are probably other classes swallowing InterruptedExceptions as well; I'm just reporting the one I found while analyzing what causes our app to hang.

For background info on thread interruption, see e.g.: http://www.ibm.com/developerworks/library/j-jtp05236/

michaelklishin:
@vfelberg yes, you are absolutely correct. By the way, you can submit a pull request that does the changes (against the stable branch), we will evaluate and merge it :)

vfelberg:
Ok, I'll make the fix; it's an easy one :-)

michaelklishin:
I'd like to get this into `3.5.2`, so we'll wait for a PR (or a note that we should handle it in our own).

vfelberg:
Unfortunately, it's not so easy to fix as I thought. I found in total 14 places catching InterruptedException; some of them trivial, most of them not. I guess, some of the core developers need to have a look at those places.

michaelklishin:
@vfelberg I will.

michaelklishin:
@vfelberg can you please reproduce the case when you have your app hanging on shutdown and post a thread dump here?

michaelklishin:
@vfelberg can you please [build from source](http://www.rabbitmq.com/build-java-client.html) on the `rabbitmq-java-client-52` branch and give this updated version a try? See `bin/dist` after `ant dist` finishes. Thank you.

PR #224, opened 2016-12-09
Handle interrupts correctly, switch to a monotonically growing clock function in BlockingCell and SingleShotLinearTimer

dhakimTRX:
Trivial threading changes in BlockingCell and SingleShotLinearTimer java utility classes
BlockingCell now consistently uses System.nanoTime rather than System.currentTimeMillis

Relates to #52

### DIFF
--- a/src/main/java/com/rabbitmq/utility/BlockingCell.java
+++ b/src/main/java/com/rabbitmq/utility/BlockingCell.java
@@ -83,12 +83,19 @@ public class BlockingCell<T> {
      * @return the waited-for value
      */
     public synchronized T uninterruptibleGet() {
-        while (true) {
-            try {
-                return get();
-            } catch (InterruptedException ex) {
-                // no special handling necessary
+        boolean wasInterrupted = false;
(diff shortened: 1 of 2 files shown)


michaelklishin:
This mashes together two orthogonal changes one of which doesn't seem to be related to #52 to me.

Can you please clarify what's the point of switching to nanoseconds in `BlockingCell`? The way blocking cells are used (for certain protocol operation continuations) does not require that kind of precision. I am not sure I understand the goal.

dhakimTRX:
BlockingCell's interrupt changes relate to #52, the others don't.  

Use of System.currentTimeMillis() causes issues whenever the system clock changes, nano time is monotonically increasing.  If my android time synced and went 2 minutes back in time during a 10 second timeout call, that call would have waited for 2 minutes, rather than timing out in 10 seconds.

michaelklishin:
@dhakimTRX thanks for the clarification!

acogoluegnes:
Thank you!

PR #227, opened 2016-12-15
Remove Default NullTrustManager and address SSL-Vulnerability Security flaw 

harrydevnull:
I am quoting the exact wordings from blog.
`. Specifically, the implementation ignores all SSL certificate validation errors when establishing an HTTPS connection to a remote host, thereby making your app vulnerable to man-in-the-middle attacks. An attacker could read transmitted data (such as login credentials) and even change the  data transmitted on the HTTPS connection.`

here instead of HTTPS it would be "TLS" 

http://www.appcelerator.com/blog/2016/02/google-security-alert-unsafe-implementation-of-the-interface-x509trustmanager/


michaelklishin:
See the response in #226.

michaelklishin:
We appreciate your taste for good security but it is not a "security flaw", please stop trying to frame it that way (and ignoring our responses, both here and on rabbitmq-users). There is a way to provide any trust manager the user pleases and there are valid scenarios where peer verification is not essential, in particular in integration tests and development environments.

PR #232, opened 2016-12-21
Close InputStream of version.properties in ClientVersion after reading

tknell:
May not break anything important, but breaks when run on Android with StrictMode enabled - which I would rather not disable.

Sorry, but I could not run the tests after the change, as I could not get the test setup (not event the build itself) running, because of missing tls-certs when running the rabbitmq-java-client, or missing makefiles in dependencies when trying to run stuff in the public-umbrella. So it would be nice if someone would run the tests for the change.

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/ClientVersion.java
+++ b/src/main/java/com/rabbitmq/client/impl/ClientVersion.java
@@ -36,7 +36,7 @@ public class ClientVersion {
         } catch (IOException e) {
         } finally {
             try {
-                if(inputStream!=null) {
+                if(inputStream != null) {
                     inputStream.close();
                 }
             } catch (IOException e) {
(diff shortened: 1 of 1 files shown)


michaelklishin:
@tknell thank you, we will test this some time this week.

michaelklishin:
This looks safe to backport to `3.6.7`.

src/main/java/com/rabbitmq/client/impl/ClientVersion.java
```diff
@@ -28,10 +29,18 @@
...
+        InputStream inputStream = ClientVersion.class.getClassLoader()
+                .getResourceAsStream("version.properties");
         try {
-            version.load(ClientVersion.class.getClassLoader()
-              .getResourceAsStream("version.properties"));
+            version.load(inputStream);
         } catch (IOException e) {
+        } finally {
+            try {
+                if(inputStream!=null) {
```
michaelklishin: Can you please add spaces around the `!=`?

tknell: Fixed. Sorry, I had to disable my formatter as it formatted things differently :) And thanks for the quick replies!

michaelklishin:
I will cherry-pick to `3.6.x`. Thank you, @tknell!

PR #244, opened 2017-02-23
Remove QueueingConsumer from RpcServer

### LINKED ISSUE #221: Remove QueueingConsumer from RpcServer
`QueueingConsumer` will be removed in 5.0.

acogoluegnes:
Fixes #221

### DIFF
--- a/src/test/java/com/rabbitmq/client/test/TestUtils.java
+++ b/src/test/java/com/rabbitmq/client/test/TestUtils.java
@@ -15,7 +15,10 @@
 
 package com.rabbitmq.client.test;
 
+import com.rabbitmq.client.Connection;
 import com.rabbitmq.client.ConnectionFactory;
+
+import java.io.IOException;
 
 public class TestUtils {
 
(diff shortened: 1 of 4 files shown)


src/test/java/com/rabbitmq/client/test/TestUtils.java
```diff
@@ -31,4 +34,14 @@ public static ConnectionFactory connectionFactory() {
         return connectionFactory;
     }
 
+    public static void close(Connection connection) {
+        if(connection != null) {
+            try {
+                connection.close();
+            } catch (IOException e) {
+                throw new RuntimeException(e);
```
michaelklishin: If we are simply ensuring that a connection is closed, does it make sense to throw a `RuntimeException` or should it be ignored? In the latter case we could rename the method to `ensureClosed`.

acogoluegnes: The initial goal was to avoid a null check in tests, and save a couple of lines. I'd prefer to know if something goes wrong in test destroy methods, instead of silently drop errors.

PR #245, opened 2017-02-23
Replace AssertionErrors with RuntimeExceptions

### LINKED ISSUE #239: Replace AssertionErrors with RuntimeExceptions
Use e.g. `IllegalStateException` instead of `AssertionError` in `BlockingCell` and other part of the code (when appropriate). The `Error` family is too dramatic for this class and client code that tries to handle exceptions from `BlockingCell` could end up dealing with cases it cannot recover from anyway.

Related to #237.

Reverse also the changes from #237, so the NIO loop doesn't actually handle `AssertionError` coming from elsewhere than `BlockingCell`.

acogoluegnes:
References #237

Fixes #239

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/nio/NioLoop.java
+++ b/src/main/java/com/rabbitmq/client/impl/nio/NioLoop.java
@@ -78,9 +78,6 @@ public class NioLoop implements Runnable {
                                 state.getConnection().handleHeartbeatFailure();
                             } catch (Exception e) {
                                 LOGGER.warn("Error after heartbeat failure of connection {}", state.getConnection());
-                            } catch (AssertionError e) {
-                                // see https://github.com/rabbitmq/rabbitmq-java-client/issues/237
-                                LOGGER.warn("Assertion error after heartbeat failure of connection {}", state.getConnection());
                             } finally {
                                 selectionKey.cancel();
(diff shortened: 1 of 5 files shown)


src/main/java/com/rabbitmq/client/impl/nio/NioLoop.java
```diff
@@ -276,17 +273,7 @@ protected void dispatchIoErrorToConnection(final SocketChannelFrameHandlerState
...
-            @Override
-            public void run() {
-                try {
-                    state.getConnection().handleIoError(ex);
-                } catch(AssertionError e) {
-                    LOGGER.warn("Assertion error during error dispatching to connection: " + e.getMessage());
-                }
-            }
-        };
+        Runnable shutdown = () -> state.getConnection().handleIoError(ex);
```
michaelklishin: Is it time to switch to lambdas and thus require JDK 8? Will Spring-AMQP require JDK 8 in 2.0?

acogoluegnes: Version 5.0 on master already requires Java 8. Spring AMQP will require Java 8 indeed: https://github.com/spring-projects/spring-amqp/blob/master/build.gradle#L82

Issue #246, opened 2017-02-23
Make it possible to use lambdas instead of *Listener objects


michaelklishin:
Now that master can require JDK 8 we can begin retrofitting the API to accept lambdas. One relatively easy step would be to make sure we accept lambdas for all the listeners which are typically SMIs (Single Method Interfaces).

For backwards compatibility we should still accept SMIs but wrap them into lambdas under the hood.

PR #251, opened 2017-02-27
Use lambdas for *Listener objects

acogoluegnes:
A proposal to use lambdas in listeners objects. Some callback
interfaces have been introduced when necessary (not necessary
for e.g. ShutdownListener). Lambdas are wrapped into the original
listener interfaces. Provide a simple test to experiment with the
API.

References #246


src/main/java/com/rabbitmq/client/Connection.java
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2007-Present Pivotal Software, Inc.  All rights reserved.
+// Copyright (c) 2007-2017 Pivotal Software, Inc.  All rights reserved.
```
michaelklishin: I think we should leave it as `Present`.

acogoluegnes:
@michaelklishin Thanks. Do you see anything to add/change? I was only thinking adding some Javadoc.

michaelklishin:
We can add a few tests that use lambdas, that's it.

PR #252, opened 2017-02-28
Add lambda-based methods to consume messages

### LINKED ISSUE #247: Minimalistic lambda-oriented method for consuming messages
Now that master can require JDK 8 we can begin retrofitting the API to accept lambdas. For example, in some clients it's common to use lambdas as consumer delivery handlers. This is generally straightforward except for one case: cancellation handling and a single lambda
wouldn't work.

I propose that we add a `Channel#basicConsume` overload that accepts two lambdas instead of a `Consumer` instance:

 * A delivery handler
 * A cancellation handler

acogoluegnes:
Only basic.deliver and basic.cancel callbacks are covered.
They're likely to be the most commonly used. This can provide
a straightforward syntax when callbacks are simple (e.g. one line).
The Delivery wrapper class from RpcServer has been extracted as
a top level class: it's used in the deliver callback to avoid
having a lambda with 4 parameters.

Fixes #247


src/main/java/com/rabbitmq/client/Delivery.java
```diff
@@ -0,0 +1,55 @@
...
+        return _properties;
+    }
+
+    /**
+     * Retrieve the message body.
+     * @return the message body
+     */
+    public byte[] getBody() {
+        return _body;
+    }
```
jplock: Can `equals` and `hashCode` be defined in here to support mocking?  The same question could be said for https://github.com/rabbitmq/rabbitmq-java-client/blob/master/src/main/java/com/rabbitmq/client/Return.java which was recently added.

jplock: Sounds good, I'll try and put one together.

PR #256, opened 2017-03-03
Reverse lock order to avoid deadlock

roelrymenants:
In class AutorecoveringConnection:

maybeDeleteRecordedAutoDeleteExchange takes the locks on
- recordedExchanges
- consumers

maybeDeleteRecordedAutoDeleteQueue takes the locks on
- recordedQueues
- consumers

Since the latter one also calls the former, the following deadlock may occur

Thread1:
maybeDeleteRecordedAutoDeleteExchange
- locked recordedExchanges
- waiting for consumers (locked by thread2)

Thread2:
maybeDeleteRecordedAutoDeleteQueue
- locked recordedQueues
- locked consumers
maybeDeleteRecordedAutoDeleteExchange
- waiting for recordedExchanges (locked by thread1)

By reversing the locks, both flows will first lock consumers, avoiding taking other locks
in a different order


michaelklishin:
Thank you, we are discussing it with @acogoluegnes. Curiously I see that it has been addressed in master but uses a different order.

PR #257, opened 2017-03-07
Add CompletableFuture async RPC to channel

### LINKED ISSUE #215: Add callback-based method in Channel
E.g. in `Channel`:
```java
void asyncRpc(Method method, CommandCallback callback) throws IOException;
```

With:
```java
public interface CommandCallback {
	void handle(Command response);
}
```

This could allow to build a reactive API and also, combined with NIO, to have a truly non-blocking client (right now RPC calls are serialized, at least at the channel level).

acogoluegnes:
Add `CompletableFuture<Command> asyncCompletableRpc(Method method)` to `Channel`. This is foundation to react to asynchronous AMQP methods (most of the resource management methods). Right now, the RPC calls are serialized at the channel level, but this could be improved in the future by using a queue for the outstanding RPC calls.

Fixes #215

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/AMQChannel.java
+++ b/src/main/java/com/rabbitmq/client/impl/AMQChannel.java
@@ -17,6 +17,7 @@
 package com.rabbitmq.client.impl;
 
 import java.io.IOException;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.TimeoutException;
 
 import com.rabbitmq.client.*;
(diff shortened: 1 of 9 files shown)


michaelklishin:
@acogoluegnes is this ready for QA? It was submitted in March but there is no "DO NOT MERGE" in the title.

acogoluegnes:
@michaelklishin Yes, it's ready for QA.

src/main/java/com/rabbitmq/client/impl/AMQChannel.java
```diff
@@ -135,6 +137,20 @@ public AMQCommand exnWrappingRpc(Method m)
         }
     }
 
+    public CompletableFuture<Command> exnWrappingAsyncRpc(Method m)
```
michaelklishin: Since this is a public method, I'd like to clarify the name. How about `exceptionWrappingAsyncRpc`?

michaelklishin: I now see we already have a "non-async" method with a similar name from way back. Disregard the above.

acogoluegnes:
@ofavre If you mean publishing a message asynchronously, `basic.publish` is already asynchronous.

Please move the discussion to [rabbitmq-users](https://groups.google.com/forum/#!forum/rabbitmq-users) if you have further questions.

PR #278, opened 2017-05-25
basicCancel and basicConsume honor rpc timeout.

vikinghawk:
Also make sure to remove other e2e bindings for autodeleted exchanges.

See https://groups.google.com/forum/#!topic/rabbitmq-users/pQ46aq6Tf_o for more details related to this PR.

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/ChannelN.java
+++ b/src/main/java/com/rabbitmq/client/impl/ChannelN.java
@@ -1239,18 +1239,26 @@ public class ChannelN extends AMQChannel implements com.rabbitmq.client.Channel 
             }
         };
 
-        rpc(new Basic.Consume.Builder()
-             .queue(queue)
-             .consumerTag(consumerTag)
-             .noLocal(noLocal)
-             .noAck(autoAck)
-             .exclusive(exclusive)
-             .arguments(arguments)
-            .build(),
-            k);
+        final Method m = new Basic.Consume.Builder()
+                .queue(queue)
+                .consumerTag(consumerTag)
+                .noLocal(noLocal)
+                .noAck(autoAck)
+                .exclusive(exclusive)
+                .arguments(arguments)
+               .build();
+        rpc(m, k);
 
         try {
-            return k.getReply();
+            if(_rpcTimeout == NO_RPC_TIMEOUT) {
+                return k.getReply();
+            } else {
+                try {
+                    return k.getReply(_rpcTimeout);
+                } catch (TimeoutException e) {
+                    throw wrapTimeoutException(m, e);
+                }
+            }
         } catch(ShutdownSignalException ex) {
             throw wrap(ex);
         }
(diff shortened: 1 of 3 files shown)


src/main/java/com/rabbitmq/client/impl/ChannelN.java
```diff
@@ -1267,17 +1275,26 @@ public void basicCancel(final String consumerTag)
         BlockingRpcContinuation<Consumer> k = new BlockingRpcContinuation<Consumer>() {
             @Override
             public Consumer transformReply(AMQCommand replyCommand) {
-                replyCommand.getMethod();
+                ((Basic.CancelOk) replyCommand.getMethod()).getConsumerTag(); // just to make sure its the method expected
```
vikinghawk: I'll be logging another issue related to this. During cluster node failovers we have seen some TimeoutExceptions during connection recovery end up causing ClassCastExceptions because the reply for the timed out  rpc request comes in while a 2nd rpc request is waiting. Adding this here to make sure the replyCommand is actually a CancelOk event. I'm not sure there is an easy fix for that so I didn't try to tackle it here.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -805,7 +805,7 @@ void maybeDeleteRecordedAutoDeleteExchange(String exchange) {
                     // last binding where this exchange is the source is gone, remove recorded exchange
                     // if it is auto-deleted. See bug 26364.
                     if((x != null) && x.isAutoDelete()) {
-                        this.recordedExchanges.remove(exchange);
+                        deleteRecordedExchange(exchange);
```
vikinghawk: It was possible to have abandoned e2e bindings in the recordedBindings.

Workflow to reproduce:
1) create durable topic exchange -> auto-delete headers exchange -> auto-delete queue -> consumer
2) cancel consumer
3) calls maybeDeleteRecordedAutoDeleteQueue which ends up deleting the recorded queue
4) calls maybeDeleteRecordedAutoDeleteExchange and removes the recorded AD headers exchange

In that case the e2e binding from the durable topic exchange to the now deleted headers exchange still existed.

acogoluegnes: This isn't directly related to the issue you fix here. Could you address this one in a separate PR? It's better for versioning, changelog, etc. Thanks :)

src/main/java/com/rabbitmq/client/impl/ChannelN.java
acogoluegnes: OK, so you want to trigger the "unsynchronized" RPC responses problem here, right? If so, maybe wrap it in try/catch block and log the exception as a warning, to avoid changing the behavior to much. Thanks.

vikinghawk: How about:

{code}
public Consumer transformReply(AMQCommand replyCommand) {
                if (!(replyCommand.getMethod() instanceof Basic.CancelOk))
                    LOGGER.warn("Received reply was not of expected method Basic.CancelOk");
                _consumers.remove(consumerTag); //may already have been removed
                dispatcher.handleCancelOk(originalConsumer, consumerTag);
                return originalConsumer;
            }
{code}

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
vikinghawk: will do

vikinghawk: https://github.com/rabbitmq/rabbitmq-java-client/pull/280

src/main/java/com/rabbitmq/client/impl/ChannelN.java
acogoluegnes: LGTM. Thanks!

acogoluegnes:
@vikinghawk Thanks!

PR #282, opened 2017-05-26
replace Error with RuntimeExceptions

vikinghawk:


vikinghawk:
Not sure why its showing the whole file changed on the compare. Must be EOL related.

Here is a link with whitespace changes ignored:
https://github.com/rabbitmq/rabbitmq-java-client/pull/282/files?w=1

vikinghawk:
oh ya i didn't catch that. Would it be better to keep the same UnexpectedMethodError class name but just change it to extend RuntimeException instead of Error?

michaelklishin:
@vikinghawk ideally we should keep as much as possible the same, except for the base class.

PR #292, opened 2017-07-07
Add option to ensure RPC reply is for the current request

### LINKED ISSUE #290: RPC timeouts can cause subsequent ClassCastExceptions

acogoluegnes:
During cluster node failovers some TimeoutExceptions during connection recovery
end up causing ClassCastExceptions because the reply for the timed out
RPC request comes in while a 2nd RPC request is waiting.

This commit add the ConnectionFactory#channelCheckRpcReplyType flag that, if enabled,
will make the channel check if a reply is compatible with the outstanding
RPC request. The flag is set to false by default but could be set to true in the future,
when the check logic is proven reliable.

This code was originally contributed by @vikinghawk.

Fixes #290

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/AMQChannel.java
+++ b/src/main/java/com/rabbitmq/client/impl/AMQChannel.java
@@ -85,7 +85,7 @@ public abstract class AMQChannel extends ShutdownNotifierComponent {
             throw new IllegalArgumentException("Continuation timeout on RPC calls cannot be less than 0");
         }
         this._rpcTimeout = connection.getChannelRpcTimeout();
-        this._checkRpcReplyType = connection.isChannelCheckRpcReplyType();
+        this._checkRpcReplyType = connection.willCheckRpcResponseType();
     }
 
     /**
--- a/src/main/java/com/rabbitmq/client/impl/AMQChannel.java
+++ b/src/main/java/com/rabbitmq/client/impl/AMQChannel.java
@@ -20,6 +20,11 @@ import java.io.IOException;
 import java.util.concurrent.TimeoutException;
 
 import com.rabbitmq.client.*;
+import com.rabbitmq.client.AMQP.Basic;
+import com.rabbitmq.client.AMQP.Confirm;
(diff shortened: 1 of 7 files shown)


vikinghawk:
+1

src/main/java/com/rabbitmq/client/impl/AMQChannel.java
```diff
@@ -153,8 +161,19 @@ public void handleCompleteInboundCommand(AMQCommand command) throws IOException
...
+                    // check if this reply command is intended for the current waiting request before calling nextOutstandingRpc()
+                    if (!_activeRpc.canHandleReply(command)) {
+                        // this reply command is not intended for the current waiting request
+                        // most likely a previous request timed out and this command is the reply for that.
+                        // Throw this reply command away so we don't stop the current request from waiting for its reply
+                        return;
+                    }
+                }
+            }
+            final RpcContinuation nextOutstandingRpc = nextOutstandingRpc();
```
vikinghawk: its a super small edge case something could change... but to be safe should probably call nextOutstandingRpc under original channelMutex lock instead of releasing and reaquiring it.

michaelklishin: This is merged already but feel free to submit a new PR.

michaelklishin:
This needs a manual merge into master because of JDK 8-oriented continuation changes.

acogoluegnes:
Merge into master done.

PR #301, opened 2017-09-14
It provides the user the flexibility to set consumer worker thread count.

2sbsbsb:
The default remains the same which is availableProcessors * 2

Background - I am using this rabbitMq java client and then found out that for each connection I have 24 worker thread and after debugging found that is availableProcessors * 2 but there is no parameters to set. 
so created this variable consumerWorkerCount by which the user will have the flexibility to control the number of worker threads.

### DIFF
--- a/src/main/java/com/rabbitmq/client/ConnectionFactory.java
+++ b/src/main/java/com/rabbitmq/client/ConnectionFactory.java
@@ -85,6 +85,9 @@ public class ConnectionFactory implements Cloneable {
     private static final String PREFERRED_TLS_PROTOCOL = "TLSv1.2";
 
     private static final String FALLBACK_TLS_PROTOCOL = "TLSv1";
+    
+    /** Default consumer worker count per connection */ 
+    public static final int DEFAULT_CONSUMER_WORKER_COUNT = Runtime.getRuntime().availableProcessors() * 2;
 
     private String username                       = DEFAULT_USER;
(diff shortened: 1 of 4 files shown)


acogoluegnes:
You can get the same result with `ConnectionFactory#setSharedExecutor(ExecutorService)`. So I don't think this new parameter is necessary.

2sbsbsb:
@acogoluegnes - sharedExecutor has one side effect that the user needs to maintain this executor service. Yes it is kind of workaround but not an ideal solution.

acogoluegnes:
Fair enough. Nevertheless I'm not sure it's worth adding yet another parameter to `ConnectionFactory`. Applications often use some container like Spring that can handle graceful shutdown for resources, including `ExecutorService`s.

WDYT @michaelklishin ?

michaelklishin:
Just provide a custom executor.

michaelklishin:
We have quite a few small concurrency-related configuration things that all can be covered with a custom executor. We also have quite a few settings on `ConnectionFactory` already and would really like to avoid adding more where possible. This is one of those cases.

2sbsbsb:
I would have liked a way that user just have a say in the no of consumer thread without complexity of maintaining it.

However I see the point made here that it is going to add a optional settings to ConnctionFactory. 
Fair enough, I will go ahead and use *shared custom execution service.

Issue #308, opened 2017-09-27
Make auto-recovery retry logic pluggable


acogoluegnes:
E.g. to implement exponential back-off. See https://groups.google.com/d/topic/rabbitmq-users/ODPEQCoyq44/discussion.

PR #309, opened 2017-09-27
Exponential backoff for recovery retry interval

vikinghawk:
Isssue #308 

First stab at a pluggable recovery retry interval and basic fibonnaci backoff implementation. Let me know your thoughts and i can make changes and add javadoc if this seems like something you guys like.

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
@@ -67,20 +67,25 @@ public class AutorecoveringChannel implements RecoverableChannel {
         try {
             delegate.close();
         } finally {
-            for (String consumerTag : consumerTags) {
-                this.connection.deleteRecordedConsumer(consumerTag);
-            }
-            this.connection.unregisterChannel(this);
+            recoveryCleanup();
         }
     }
 
     @Override
     public void close(int closeCode, String closeMessage) throws IOException, TimeoutException {
         try {
-          delegate.close(closeCode, closeMessage);
+            delegate.close(closeCode, closeMessage);
         } finally {
-          this.connection.unregisterChannel(this);
+            recoveryCleanup();
         }
+    }
+    
+    private void recoveryCleanup() {
+        for (String consumerTag : consumerTags) {
+            this.connection.deleteRecordedConsumer(consumerTag);
+        }
+        // TODO what about any other recorded queues and bindings that were owned by this channel? They will now cause recovery exceptions.
+        this.connection.unregisterChannel(this);
     }
 
     @Override
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
@@ -67,25 +67,20 @@ public class AutorecoveringChannel implements RecoverableChannel {
         try {
             delegate.close();
         } finally {
-            recoveryCleanup();
+            for (String consumerTag : consumerTags) {
+                this.connection.deleteRecordedConsumer(consumerTag);
+            }
+            this.connection.unregisterChannel(this);
         }
     }
 
     @Override
     public void close(int closeCode, String closeMessage) throws IOException, TimeoutException {
         try {
-            delegate.close(closeCode, closeMessage);
+          delegate.close(closeCode, closeMessage);
         } finally {
-            recoveryCleanup();
+          this.connection.unregisterChannel(this);
         }
-    }
-    
-    private void recoveryCleanup() {
-        for (String consumerTag : consumerTags) {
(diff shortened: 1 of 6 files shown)


pivotal-issuemaster:
@vikinghawk Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-java-client&pullRequestId=309)!

[Click here](https://cla.pivotal.io/sync/rabbitmq?repositoryId=rabbitmq/rabbitmq-java-client&pullRequestId=309) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@vikinghawk Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/rabbitmq?repositoryId=rabbitmq/rabbitmq-java-client&pullRequestId=309)!

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
```diff
@@ -67,21 +67,26 @@ public void close() throws IOException, TimeoutException {
...
-          this.connection.unregisterChannel(this);
+            recoveryCleanup();
         }
     }
+    
+    private void recoveryCleanup() {
+        for (String consumerTag : consumerTags) {
+            this.connection.deleteRecordedConsumer(consumerTag);
+        }
+        // TODO what about any other recorded queues and bindings that were owned by this channel? They will now cause recovery exceptions.
```
michaelklishin: Queues and bindings are not "owned" by channels in any way. Consumers are channel state, however. Queues and bindings are tracked by connections (in part because queues can be owned by connections — when they are exclusive — and in part because declaring something on one channel then deleting it on another should leave no traces of that entity.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
```diff
@@ -67,21 +67,26 @@ public void close() throws IOException, TimeoutException {
         try {
             delegate.close();
         } finally {
-            for (String consumerTag : consumerTags) {
-                this.connection.deleteRecordedConsumer(consumerTag);
-            }
-            this.connection.unregisterChannel(this);
+            recoveryCleanup();
```
michaelklishin: Is this change really related to delay calculation and back-off?

michaelklishin:
In general this looks promising.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
vikinghawk: no, just noticed the recorded consumers weren't being deleted with the 2nd close method like the 1st.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
vikinghawk: Ya thats a bad description. What that is referencing is that any queues or bindings that were created with this channel still have the recorded entities reference this now closed channel and will throw an exception during recovery. I don't know a good way to address this other than having recorded queues/exchanges/bindings use a special "recovery" channel instead of the original creating channel. I assume the only recorded items that would require to be tied to the original channel that created them would be consumers which are cleaned up.

vikinghawk: removed this change.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringChannel.java
vikinghawk: removed for now. could be a separate future PR

vikinghawk:
Updated with javadoc, test, and suggested changes.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -541,8 +542,7 @@ private RecoveryAwareAMQConnection recoverConnection() throws InterruptedExcepti
 				newConn.abort();
 				return null;
             } catch (Exception e) {
-                // TODO: exponential back-off
-                Thread.sleep(this.params.getNetworkRecoveryInterval());
+                Thread.sleep(this.params.getRecoveryDelayHandler().getDelay(attempts));
```
vikinghawk: thoughts on flipping the order of this? so deliver the exception before sleeping?

acogoluegnes: Let's keep it the way it is.

michaelklishin: Good catch. I think at some point we will have to do it. The way it works right now is a bit weird and can complicate additional manual recovery steps.

PR #329, opened 2017-11-07
Add property file-based initialization

### LINKED ISSUE #324: property file-based initialization
This is a feature request for the RabbitMQ Java library.

It would be useful if we could initialize the `ConnectionFactory` via a properties file. Now, whenever we want to set a new configuration property, we have to change the code to call the setter. It would be more convenient if we just could alter the properties file.

You could take Hikari as an example, see https://github.com/brettwooldridge/HikariCP#initialization. A java.util.properties file can be provided. Whenever we want to alter the configuration, we just update the properties file, without being forced to alter the code.

```
// Examines both filesystem and classpath for .properties file
HikariConfig config = new HikariConfig("/some/path/hikari.properties");
HikariDataSource ds = new HikariDataSource(config);
```

acogoluegnes:
Fixes #324


src/main/java/com/rabbitmq/client/ConnectionFactory.java
```diff
@@ -1060,6 +1060,74 @@ protected AddressResolver createAddressResolver(List<Address> addresses) {
...
     /**
+     * Load settings from a property file.
+     * The default prefix for keys is <code>rabbitmq.</code>
+     * @param propertyFileLocation location of the property file to use
+     * @throws IOException when something goes wrong reading the file
+     * @since 4.4.0
+     * @see ConnectionFactoryConfigurer
+     */
+    public void load(String propertyFileLocation) throws IOException {
+        ConnectionFactoryConfigurer.load(this, propertyFileLocation);
```
michaelklishin: I think `Configurator` is a more common word in English.

src/main/java/com/rabbitmq/client/ConnectionFactory.java
```diff
@@ -1060,6 +1060,74 @@ protected AddressResolver createAddressResolver(List<Address> addresses) {
...
+    }
+
+    /**
+     * Load settings from a property file.
+     * @param propertyFileLocation location of the property file to use
+     * @param prefix prefix used in the property file keys
```
michaelklishin: It's not really clear from this line how the prefix is used.

michaelklishin: Perhaps a "see ConnectionFactoryConfigurer" link would help the most here.

src/main/java/com/rabbitmq/client/ConnectionFactoryConfigurer.java
```diff
@@ -0,0 +1,252 @@
...
+    public static final String DEFAULT_PREFIX = "rabbitmq.";
+
+    public static final String USERNAME = "username";
+    public static final String PASSWORD = "password";
+    public static final String VIRTUAL_HOST = "virtual.host";
+    public static final String HOST = "host";
+    public static final String PORT = "port";
+    public static final String REQUESTED_CHANNEL_MAX = "requested.channel.max";
+    public static final String REQUESTED_FRAME_MAX = "requested.frame.max";
+    public static final String REQUESTED_HEARTBEAT = "requested.heartbeat";
```
michaelklishin: I'd use "heartbeat.timeout" here. Same for `frame.max` and `channel.max`.

src/main/java/com/rabbitmq/client/ConnectionFactoryConfigurer.java
```diff
@@ -0,0 +1,252 @@
...
+ * @since 4.4.0
+ * @see ConnectionFactory#load(String, String)
+ */
+public class ConnectionFactoryConfigurer {
+
+    public static final String DEFAULT_PREFIX = "rabbitmq.";
+
+    public static final String USERNAME = "username";
+    public static final String PASSWORD = "password";
+    public static final String VIRTUAL_HOST = "virtual.host";
```
michaelklishin: `vhost`?

acogoluegnes: Property name is `virtualHost` in `ConnectionFactory`, so I'm trying to be consistent. WDYT?

src/main/java/com/rabbitmq/client/ConnectionFactoryConfigurer.java
acogoluegnes: Again, I tried to map with the existing properties in `ConnectionFactory`. Consistency over better naming?  :-)

michaelklishin: It's just weird that we use dots so aggressively. In case of `requested.heartbeat` we propagate an unfortunately named property and make it even more confusing by using dot separators.

Can we use `connection.channel_max`, `connection.frame_max`, `connection.heartbeat.timeout` for those 3?

acogoluegnes: Sure.

src/main/java/com/rabbitmq/client/ConnectionFactoryConfigurator.java
```diff
@@ -0,0 +1,252 @@
...
+    public static final String PORT = "port";
+    public static final String CONNECTION_CHANNEL_MAX = "connection.channel.max";
+    public static final String CONNECTION_FRAME_MAX = "connection.frame.max";
+    public static final String CONNECTION_HEARTBEAT = "connection.heartbeat";
+    public static final String CONNECTION_TIMEOUT = "connection.timeout";
+    public static final String HANDSHAKE_TIMEOUT = "handshake.timeout";
+    public static final String SHUTDOWN_TIMEOUT = "shutdown.timeout";
+    public static final String USE_DEFAULT_CLIENT_PROPERTIES = "use.default.client.properties";
+    public static final String CLIENT_PROPERTIES_PREFIX = "client.properties.";
+    public static final String AUTOMATIC_RECOVERY_ENABLED = "automatic.recovery.enabled";
```
michaelklishin: This can probably be `connection.recovery.enabled`. We already have `topology.recovery.enabled`.

src/main/java/com/rabbitmq/client/ConnectionFactoryConfigurator.java
```diff
@@ -0,0 +1,252 @@
...
+    public static final String TOPOLOGY_RECOVERY_ENABLED = "topology.recovery.enabled";
+    public static final String NETWORK_RECOVERY_INTERVAL = "network.recovery.interval";
```
michaelklishin: `connection.recovery.interval`?

PR #345, opened 2018-01-17
Proposal: Add method to remove a queue from recovery without deleting it

vikinghawk:
## Proposed Changes
This is a proposal to add a new method to address the following usecase:
1. Connection A declares a durable, shared,non-autodelete queue "ABC"
2. Connection A consumes from the queue
3. Connection B re-declares the same queue "ABC" as it does not know or care whether it already exists from Connection A
4. Connection B consume from the queue
5. Connection A cancels its queue consumers
6. Connection A does not want to delete the queue because it may still be in use by others, but it would like to remove the queue from its list of recovery items because it is no longer actively using it.

This adds a public method to AutorecoveringConnection to allow for removal of a queue from recorded queues (and its related bindings) without having to call Channel.queueDelete().

Let me know if this is something you guys would be willing to pull in. I was originally looking at using reflections to get to the necessary methods/variables to do this. But obviously that is fragile for me to maintain and there may be others in the community with a similar usecase that could benefit.

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
@@ -761,6 +761,28 @@ public class AutorecoveringConnection implements RecoverableConnection, NetworkC
             this.maybeDeleteRecordedAutoDeleteExchange(b.getSource());
         }
     }
+    
+    /**
+     * Remove the queue from list of recorded queues to recover after connection failure.
+     * Intended to be used in usecases where you want to remove the queue from this connection's recovery list but don't want to delete the queue from the server.
+     * 
+     * @param queue queue name to remove from recorded recovery queues
+     * @param ifUnused if true, the RecordedQueue will only be removed if no more consumers from this connection are using it.
+     */
+    public void removeRecordedQueue(final String queue, final boolean ifUnused) {
+        if (ifUnused) {
+            // Note: This is basically the same as maybeDeleteRecordedAutoDeleteQueue except it works for non auto-delete queues as well.
+            synchronized (this.consumers) {
+                synchronized (this.recordedQueues) {
+                    if (!hasMoreConsumersOnQueue(this.consumers.values(), queue)) {
+                        deleteRecordedQueue(queue);
+                    }
+                }
+            }
+        } else {
+            deleteRecordedQueue(queue);
+        }
+    }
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -761,6 +761,28 @@ void deleteRecordedQueue(String queue) {
...
     }
+    
+    /**
+     * Remove the queue from list of recorded queues to recover after connection failure.
+     * Intended to be used in usecases where you want to remove the queue from this connection's recovery list but don't want to delete the queue from the server.
+     * 
+     * @param queue queue name to remove from recorded recovery queues
+     * @param ifUnused if true, the RecordedQueue will only be removed if no more consumers from this connection are using it.
+     */
+    public void removeRecordedQueue(final String queue, final boolean ifUnused) {
```
michaelklishin: WDYT about `excludeQueueFromRecovery` or `doNotRecoveryQueue`?

michaelklishin:
The scenario sounds reasonable to me. I'm not sure the name is what we should go with (suggested a couple of alternatives above) but otherwise no objections from me @acogoluegnes.

src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
vikinghawk: I like doNotRecoverQueue or removeQueueFromRecovery.
excludeQueueFromRecovery works too. I think we would just need to make it clear you have to call the method after declaring a queue and not before. Its not adding an exclude filter or anything like that for future declareQueue calls.

michaelklishin: My personal favorite is `excludeQueueFromRecovery`.

acogoluegnes: I'd go with `excludeQueueFromRecovery`.

vikinghawk: works for me :)
I'll change that and look at adding a junit.
Thanks guys.

vikinghawk:
Changes made and moved to 4.x.x branch on #346

Issue #375, opened 2018-06-27
Continuations (requests) are not discarded on timeouts


wheezil:
I am starting to use the RpcClient class, and examining the source it apparently can leak by adding entries to _continuationMap that are never removed if there is no reply.  If doCall() throws TimeoutException, there is no reason to leave this entry in the map as it will never be used, so it should catch TimeoutException and remove the entry.

* RabbitMQ version: 5.3.0 (but issue remains in master code)

Relevant code follows.  The call to k.uninterruptibleGet() can throw TimeoutException. We can catch that and remove the entry before re-throwing it.

    public Response doCall(AMQP.BasicProperties props, byte[] message, int timeout)
        throws IOException, ShutdownSignalException, TimeoutException {
        checkConsumer();
        BlockingCell<Object> k = new BlockingCell<Object>();
        synchronized (_continuationMap) {
            _correlationId++;
            String replyId = "" + _correlationId;
            props = ((props==null) ? new AMQP.BasicProperties.Builder() : props.builder())
                .correlationId(replyId).replyTo(_replyTo).build();
            _continuationMap.put(replyId, k);
        }
        publish(props, message);
        Object reply = k.uninterruptibleGet(timeout);
        if (reply instanceof ShutdownSignalException) {
            ShutdownSignalException sig = (ShutdownSignalException) reply;
            ShutdownSignalException wrapper =
                new ShutdownSignalException(sig.isHardError(),
                                            sig.isInitiatedByApplication(),
                                            sig.getReason(),
                                            sig.getReference());
            wrapper.initCause(sig);
            throw wrapper;
        } else {
            return (Response) reply;
        }
    }

I've already posted this to the rabbitmq mailing list, and got the thumbs-up to submit a patch, so here goes!  Please forgive my newbie status; I am still struggling getting this to build.

PR #376, opened 2018-06-27
Fix RpcClient potential memory leak if timeout occurs

wheezil:
## Proposed Changes

The RpcClient can time out if no response is received within the time window. If this happens, it doesn't necessarily leak, but it *will* leak if the response is *never* received. This change recognizes that once a timeout happens, the entry in the _continuationMap serves no purpose and can be safely removed, since the client cannot wait again on the same request.

## Types of Changes

- [X] Bugfix (non-breaking change which fixes issue #375)

## Checklist

- [X] I have read the `CONTRIBUTING.md` document
- [X] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- I was not able to add a test because of trouble building.  However I did run a local test program (see https://github.com/wheezil/rabbitmq-rpc-example) to confirm that that a client (a) times out and removes the entry (b) causes no later issues when the reply arrives
- [X] I have added necessary documentation (if appropriate)
- [X] Any dependent changes have been merged and published in related repositories

## Further Comments

Around line 197, I allow for the entry to have already been removed when the reply finally arrives  and ignore its absence instead of throwing IllegalStateException.  I don't *think* this runs afoul of any assumptions, but it would be good if another pair of eyes looked this over.

### DIFF
--- a/src/main/java/com/rabbitmq/client/RpcClient.java
+++ b/src/main/java/com/rabbitmq/client/RpcClient.java
@@ -192,10 +192,10 @@ public class RpcClient {
                     String replyId = properties.getCorrelationId();
                     BlockingCell<Object> blocker =_continuationMap.remove(replyId);
                     if (blocker != null) {
-	                    blocker.set(new Response(consumerTag, envelope, properties, body));
+                        blocker.set(new Response(consumerTag, envelope, properties, body));
                     } else {
-						// Not an error. Entry will have been removed if request timed out.
-					}
+                        // Not an error. Entry will have been removed if request timed out.
+                    }
                 }
             }
         };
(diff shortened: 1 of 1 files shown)


acogoluegnes:
Thanks @wheezil for this contribution! I'll backport it to `5.3.x-stable` and will add a test.

acogoluegnes:
@wheezil I added a test case and ended up logging a warning instead of silently ignoring a missing outstanding request: https://github.com/rabbitmq/rabbitmq-java-client/commit/11fdaa9c1f0f4ea102cba49ec9c820b1092e1d61

wheezil:
That’s even better
John Lilley

PR #388, opened 2018-08-07
Add optional retry logic to topology recovery

### LINKED ISSUE #387: Add (optional) retry logic to topology recovery
Topology recovery could benefit from retry logic on failed operations for some specific cases, e.g. failed binding because auto-delete queues has been deleted during the topology recovery.

The retry logic would be:
 * optional
 * configurable to some extent: retry condition (which error, which recoverable entity), retry action, number of retries, backoff policy

acogoluegnes:
There's no topology recovery retry by default. The default
implementation is composable: not all have the recoverable entities have
to retry and the retry operations don't have to be only the
corresponding entity recovery, but also other operations, like
recovering the corresponding channel.

Fixes #387


michaelklishin:
It's not worth rejecting the review over but: should we retry at least twice by default? Currently `retryAttempts` defaults to 1. With 2 it would be 3 attempts total (including the original one).

michaelklishin:
I like how functional-ish (reusable and composable) the lambda-based approach turned out to be 👍

PR #403, opened 2018-08-28
tweaks to recovery retry

vikinghawk:
## Proposed Changes

Changed the helper method to return the builder instead of the built RetryHandler so it is easy to add your own backoff impl and chagne retry attempts. Also a tweak to the logging so it logs each retry loop and how many attempts are remaining.

## Types of Changes

## Checklist

## Further Comments

considered, etc.


src/main/java/com/rabbitmq/client/impl/recovery/DefaultRetryHandler.java
```diff
@@ -122,11 +122,11 @@ protected RetryResult doRetry(BiPredicate<RecordedEntity, Exception> condition,
                 throw exception;
             }
         }
-        throw context.exception();
+        throw exception;
```
vikinghawk: assuming we want the last exception thrown here rather than the 1st? If not i can change it back.

acogoluegnes: Makes sense.

acogoluegnes:
Thanks! Changes are minor but I'll release a RC for both 5.x and 4.x today.

PR #420, opened 2018-10-26
Add mandatory flag option to RpcClient

### LINKED ISSUE #417: An option to make RpcClient fail early if there are no bindings
I have noticed that the RpcClient utiltity has no way of using the 'mandatory' flag. 
I would like my RPC call to fail 'fast' in case nobody is listening to the request. Wihout it, i have to wait for the RPC to timeout. 

* RabbitMQ client version 5.5.0

acogoluegnes:
[#161476914]

Fixes #417

### DIFF
--- a/src/main/java/com/rabbitmq/client/RpcClient.java
+++ b/src/main/java/com/rabbitmq/client/RpcClient.java
@@ -107,7 +107,7 @@ public class RpcClient {
             throw new IllegalArgumentException("Timeout argument must be NO_TIMEOUT(-1) or non-negative.");
         }
         _timeout = params.getTimeout();
-        _useMandatory = params.isUseMandatory();
+        _useMandatory = params.shouldUseMandatory();
         _replyHandler = params.getReplyHandler();
         _correlationId = 0;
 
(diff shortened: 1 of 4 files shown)


src/main/java/com/rabbitmq/client/RpcClient.java
```diff
@@ -67,6 +89,46 @@
...
+            IOException {
+        _channel = params.getChannel();
+        _exchange = params.getExchange();
+        _routingKey = params.getRoutingKey();
+        _replyTo = params.getReplyTo();
+        if (params.getTimeout() < NO_TIMEOUT) {
+            throw new IllegalArgumentException("Timeout argument must be NO_TIMEOUT(-1) or non-negative.");
+        }
+        _timeout = params.getTimeout();
+        _useMandatory = params.isUseMandatory();
```
michaelklishin: Should this be `shouldUseMandatory`?

acogoluegnes: Makes sense. Changed.

PR #437, opened 2018-12-26
Fix warnings in ClientVersion when amqp-client is relocated

### LINKED ISSUE #436: Sprious warnings emitted by ClientVersion when the library is relocated
If the RabbitMQ library is relocated, it will emit a spurious warning about not being able to find the client version. This is because the property string (`com.rabbitmq.client.version`), is rewritten but `rabbitmq-amqp-client.properties` is not.

This is not a server issue, only a client library issue that does not affect operations, but for completeness:

* RabbitMQ version: 3.7.8
* Erlang version: `Erlang/OTP 20 [erts-9.3] [source] [64-bit] [smp:2:2] [ds:2:2:10] [async-threads:64] [hipe] [kernel-poll:true]`
* Other relevant information: https://gist.github.com/astei/1e439c19a29993d9486ff4a5afb9f145
* [Smallest possible reproduction demo](https://github.com/astei/rabbitmq-relocation-bug)
* Client library version (for all libraries used): Java client 5.5.1
* Operating system, version, and patch level: Alpine Linux 3.8.1 on both server and client (we use Kubernetes to run RabbitMQ)

I have a PR to fix this issue, which I will open shortly.

astei:
## Proposed Changes

This pull request fixes spurious warnings emitted by the `com.rabbitmq.client.impl.ClientVersion` class when `amqp-client` is relocated. This problem happens because the key `com.rabbitmq.client.version` is rewritten to use a different prefix by whatever produces the uberjar (`maven-shade-plugin` or Gradle Shadow, for instance), but the `rabbitmq-amqp-client.properties` key is not rewritten.

In order to prevent this rewriting, the property key is now stored in a `char[]` and we create a `String` from it. This is unusual, but required if we want to retain backwards-compatibility. Doing so allows us to defeat the rewriting done during relocation.

This behavior has been broken for a while: prior to bd1d73fc2e, the version returned could be `null`.

This issue fixes #436.

While we're here, I also fixed a typo in the exception message thrown by `getVersionFromPropertyFile()`.

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #436)
- [X] Cosmetics (whitespace, appearance)

## Checklist

- [X] I have read the `CONTRIBUTING.md` document
- [X] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [X] All tests pass locally with my changes
- [X] I have added necessary documentation (if appropriate)

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/ClientVersion.java
+++ b/src/main/java/com/rabbitmq/client/impl/ClientVersion.java
@@ -28,6 +28,11 @@ public class ClientVersion {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(ClientVersion.class);
 
+    // We store the version property in an unusual way because relocating the package can rewrite the key in the property
+    // file, which results in spurious warnings being emitted at start-up.
+    private static final char[] VERSION_PROPERTY = new char[] {'c', 'o', 'm', '.', 'r', 'a', 'b', 'b', 'i', 't', 'm', 'q', '.',
+            'c', 'l', 'i', 'e', 'n', 't', '.', 'v', 'e', 'r', 's', 'i', 'o', 'n'};
+
     public static final String VERSION;
 
     static {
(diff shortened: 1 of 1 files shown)


michaelklishin:
@astei may I ask you to clarify how instantiating a string from a character array addresses this? I’m afraid I haven’t used Gradle Shadow.

astei:
@michaelklishin A very good question to ask!

Gradle Shadow uses ASM for relocating classes. By default, [ASM remaps just `Class<?>` and `Handle`](https://github.com/consulo/objectweb-asm/blob/master/asm-commons/src/main/java/org/objectweb/asm/commons/Remapper.java#L152). Gradle Shadow adds [mappings for `String`](https://github.com/johnrengelman/shadow/blob/master/src/main/groovy/com/github/jengelman/gradle/plugins/shadow/impl/RelocatorRemapper.groovy#L53).

Obviously, `char[]` is not remapped in either case, so we can use it to avoid this mapping step.

Issue #626, opened 2019-09-12
Add support for OAuth 2 authentication


acogoluegnes:
RabbitMQ 3.8 ships with [OAuth 2 support](https://github.com/rabbitmq/rabbitmq-auth-backend-oauth2). The client should provide utility to retrieve JWT tokens and to refresh them when appropriate.

PR #627, opened 2019-09-16
Do not depend on generated classes for update-secret extension

acogoluegnes:
Some CI testing jobs are using the 3.7 branch of codegen, which does not contain
the update-secret extension. The Java client then cannot be built
because it requires some generated code for the extension.

This commit adds the code to handle update-secret and thus
makes the generated code not necessary. This is just a workaround for
these testing jobs to succeed, it does not change the handling of the
update-secret extension in the client.

[#167029587]

References #626


michaelklishin:
I'm not sure about the precedent this is setting. We could simply update the `v3.7.x` pipeline. But this is an acceptable solution. Thanks for the quick turnaround.

Issue #637, opened 2020-01-29
Allow configurable correlatorId generation in RpcClient


janssk1:
The RpcClient class has a hardcoded algorithm for unique correlatorId generation. This works in most cases. However, i recently hit an issue where the 'correlatorId' was used (or misused) to encode some extra information needed by the receiver. It would be great if the RpcClient could be configured with a 'strategy' to generate unqiue correlatorIds.  
Alternatively, the RpcClient should not touch the correlatorId if it is already provided in the arguments. 
* RabbitMQ client version
5.8.0

michaelklishin:
This sounds like a great opportunity to contribute. This is open-source software after all ;)

acogoluegnes:
This should be easy to add in a non-breaking way, especially now there's a `RpcClient` constructor with `RpcClientParams`. We've already done that in the [Reactor extension](https://github.com/reactor/reactor-rabbitmq/blob/97f091c44d32f10d3022b11c8824a32470f68800/src/main/java/reactor/rabbitmq/RpcClient.java#L73), with a `Supplier<String>` to create the correlation ids.

A contribution would be indeed welcome.

acogoluegnes:
Fixes in #638.

PR #638, opened 2020-02-03
637: Configurable correlatorId generation for RPCClient

janssk1:
Change-Id: I17d7a214d3336ad6e5fe890857a48b457ea599b6

## Proposed Changes

https://github.com/rabbitmq/rabbitmq-java-client/issues/637

## Types of Changes

- [x ] New feature (non-breaking change which adds functionality)
Did remove the 'getCorrelatorId' method from RpcClient. Since i don't see a use for, did not bother adding compatiblity hacks

## Checklist

- [ ] I have read the `CONTRIBUTING.md` document
- [x ] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [- ] All tests pass locally with my changes
Tried it. But gave up after a few hours. I don't have time to troubleshoot all environment issues that pop up when trying to run the 'unit' tests
- [ x] I have added tests that prove my fix is effective or that my feature works

### DIFF
--- a/src/main/java/com/rabbitmq/client/RpcClient.java
+++ b/src/main/java/com/rabbitmq/client/RpcClient.java
@@ -80,8 +80,6 @@ public class RpcClient {
         }
     };
 
-    public static Supplier<String> DEFAULT_CORRELATION_ID_GENERATOR = new IncrementingCorrelationIdGenerator("");
-
     private final Function<Object, Response> _replyHandler;
 
     /** Map from request correlation ID to continuation BlockingCell */
--- a/src/main/java/com/rabbitmq/client/RpcClient.java
+++ b/src/main/java/com/rabbitmq/client/RpcClient.java
@@ -28,6 +28,7 @@ import java.util.Map;
 import java.util.Map.Entry;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Function;
+import java.util.function.Supplier;
 
 import com.rabbitmq.client.impl.MethodArgumentReader;
 import com.rabbitmq.client.impl.MethodArgumentWriter;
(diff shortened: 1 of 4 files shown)


src/main/java/com/rabbitmq/client/RpcClient.java
```diff
@@ -79,12 +80,14 @@
         }
     };
 
+    public static Supplier<String> DEFAULT_CORRELATION_ID_GENERATOR = new IncrementingCorrelationIdGenerator("");
```
Stephan202: As-is this static instance may be used by more than one `RpcClient` instance. In that case `Supplier#get` may be invoked concurrently, as the `synchronized (_continuationMap)` block only ensures exclusive access for `doCall` invocations on the same client. This is a problem because as-is `IncrementingCorrelationIdGenerator` is not thread-safe.

One option would be to replace `int _correlationId` with an `AtomicInteger` (or perhaps more appropriately, an `AtomicLong`.)

janssk1: Good catch. This should not be static. Will remove that.

src/main/java/com/rabbitmq/client/RpcClient.java
```diff
@@ -389,14 +389,6 @@ public String getRoutingKey() {
         return _continuationMap;
     }
 
-    /**
```
acogoluegnes: This is a breaking change. The last correlation ID must be kept in some way, likely by keeping the `_correlationId` property and update it in the synchronized block.

src/main/java/com/rabbitmq/client/RpcClient.java
```diff
@@ -392,9 +392,14 @@ public String getRoutingKey() {
...
      * @return the most recently used correlation id
+     * @deprecated Only works for {@link IncrementingCorrelationIdGenerator}
      */
     public int getCorrelationId() {
-        return _correlationId;
+        if (_correlationIdGenerator instanceof IncrementingCorrelationIdGenerator) {
+            return ((IncrementingCorrelationIdGenerator) _correlationIdGenerator).getCorrelationId();
+        } else {
+            throw new UnsupportedOperationException();
+        }
```
acogoluegnes: Not super fan of this code TBH. Maybe keeping a reference to the last correlation ID (a string) and trying to convert it to an integer when `getCorrelationId()` is called would do the job as well. This is a best-effort approach anyway, but it avoids relying on some implementation.

janssk1: This is implementation is only there for BW compatiblity, So i don't think it's a problem that it only works for the 'default' id generator. I marked the method as deprecated anyway.

acogoluegnes:
Thanks!

PR #667, opened 2021-02-13
Binding recovery retry should recover all bindings on the recovered queue

vikinghawk:
## Proposed Changes
When a binding recovery fails and retry logic kicks in, the code should recreate all of the bindings on the destination and not just the binding that failed.

For a queue with a large amount of bindings, its possible that some of the first binding recoveries succeeded, but then the queue got deleted and a later binding failed. In that scenario, we need to make sure all the earlier bindings that were already recovered are created as well.

Background: We hit this scenario on a cluster that had a large amount of queues and bindings. The queue in question had an expiry policy set on it that caused the queue to get deleted before all the binding recoveries had finished. The retry logic did its job and recreated the queue and the failed bindings. But we were silently missing the 1st set of bindings that had originally passed. We made a change to the policy to increase the expiry time so this scenario shouldn't happen again. But I thought it best to implement this fix as well.

## Types of Changes

- [ x] Bugfix (non-breaking change which fixes issue #NNNN)

## Checklist

- [x ] I have read the `CONTRIBUTING.md` document

## Further Comments

I believe these tests should work... However, I am on windows and had issues getting the Host.rabbitmqctl() pieces to work. So when running locally i commented those pieces out and manually dropped the connections and queues and the tests were passing that way.

### DIFF
--- a/src/test/java/com/rabbitmq/client/test/functional/TopologyRecoveryRetry.java
+++ b/src/test/java/com/rabbitmq/client/test/functional/TopologyRecoveryRetry.java
@@ -45,7 +45,7 @@ import static org.junit.Assert.assertTrue;
  */
 public class TopologyRecoveryRetry extends BrokerTestCase {
 
-    private Consumer<Integer> backoffConsumer;
+    private volatile Consumer<Integer> backoffConsumer;
     
     @After
     public void cleanup() {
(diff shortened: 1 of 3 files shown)


src/test/java/com/rabbitmq/client/test/functional/TopologyRecoveryRetry.java
```diff
@@ -40,18 +60,149 @@ public void topologyRecoveryRetry() throws Exception {
...
+            }
+            @Override
+            public void handleRecovery(Recoverable recoverable) {
+                recoveryLatch.countDown();
+            }
+        });
+        
+        // we want recovery to fail when recovering the 2nd binding
+        // give the 2nd recorded binding a bad queue name so it fails
+        final RecordedBinding binding2 = ((AutorecoveringConnection)connection).getRecordedBindings().get(1);
```
michaelklishin, line 100: This test is brutal :D

michaelklishin:
So for every binding out of N, we will now recover N bindings, meaning N^2 `queue.bind` operations will be performed for every queue? This sounds really excessive.

vikinghawk:
This retry logic only kicks in when a binding recovery has failed due to a 404 error from the queue being deleted. And it will only recover bindings for the specific queue that was deleted, not all queues.
But yes, the logic as I had it was potentially recovering more than it needed to and could cause later bindings to be recovered twice.

I have now tweaked the logic to only recover the queue bindings that were recovered before the failed binding.

michaelklishin:
Yes, this makes more sense. @acogoluegnes any objections from you?

acogoluegnes (approved):

acogoluegnes:
Thanks!

michaelklishin:
Forward-ported to `master`.

PR #668, opened 2021-02-15
Add topology recovery started method to RecoveryListener

vikinghawk:
## Proposed Changes

This adds a default method to the RecoveryListener interface that notifies the listener when connection & channel recovery is complete and that topology recovery is now starting.

Once the connection & channels have been recovered, it is safe for an application to start publishing messages again. For connections that have a lot of queues/bindings/consumers to recover, it can be useful for applications to know this intermediary state and not have to wait for all of the subscription related entities to be fully recovered.

## Types of Changes

## Checklist

## Further Comments

considered, etc.

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
@@ -561,6 +561,10 @@ public class AutorecoveringConnection implements RecoverableConnection, NetworkC
 
     private synchronized void beginAutomaticRecovery() throws InterruptedException {
         this.wait(this.params.getRecoveryDelayHandler().getDelay(0));
+        final long delay = this.params.getRecoveryDelayHandler().getDelay(0);
+        if (delay > 0)  {
+            this.wait(delay);
+        }
 
         this.notifyRecoveryListenersStarted();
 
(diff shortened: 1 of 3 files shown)


src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -560,7 +560,10 @@ public void removeConsumerRecoveryListener(ConsumerRecoveryListener listener) {
     }
 
     private synchronized void beginAutomaticRecovery() throws InterruptedException {
-        this.wait(this.params.getRecoveryDelayHandler().getDelay(0));
+        final long delay = this.params.getRecoveryDelayHandler().getDelay(0);
+        if (delay > 0)  {
```
vikinghawk, line 564: i had tried to use a backoff recovery delay handler that had an initial delay of 0 and quickly realized that recovery would hang forever here. An easy work around is to use a 1ms initial delay, but this feels like a better solution that may save someone else in the future.

michaelklishin: This is reasonable if you ask me.

michaelklishin (approved):

michaelklishin:
Thank you!

PR #672, opened 2021-02-19
Handle basic query parameters in connection URI

captainju:
## Proposed Changes

The current connection method `setUri()` does not handle query parameters at all, unlike the official Erlang client.
This pull request adds basic query parameters capabilities.
However, due to SSL handling differences between the Erlang client and the Java client, all SSL parameters are not used.
Only the following parameters are used:

- heartbeat : Heartbeat timeout value in seconds (an integer) to negotiate with the server.
- connection_timeout : Time in milliseconds (an integer) to wait while establishing a TCP connection to the server before giving up.
- channel_max : Maximum number of channels to permit on this connection.

(taken from https://www.rabbitmq.com/uri-query-parameters.html)

It is easily extendable for other future query parameters.

## Types of Changes

## Checklist

### DIFF
--- a/src/main/java/com/rabbitmq/client/ConnectionFactory.java
+++ b/src/main/java/com/rabbitmq/client/ConnectionFactory.java
@@ -348,6 +348,11 @@ public class ConnectionFactory implements Cloneable {
 
             setVirtualHost(uriDecode(uri.getPath().substring(1)));
         }
+        
+        String rawQuery = uri.getRawQuery();
+        if (rawQuery != null && rawQuery.length() > 0) {
+            setQuery(rawQuery);
+        }
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/com/rabbitmq/client/ConnectionFactory.java
```diff
@@ -377,6 +382,65 @@ private static String uriDecode(String s) {
...
+     * @param rawQuery is the string containing the raw query parameters part from a URI
+     */
+    private void setQuery(String rawQuery) {
+        Map<String, String> parameters = new HashMap<>();
+        
+        // parsing the query parameters
+        try {
+            for (String param : rawQuery.split("&")) {
+                String[] pair = param.split("=");
+                String key = URLDecoder.decode(pair[0], "US-ASCII");
```
michaelklishin, line 399: Any reason to not use `UTF-8` here?

michaelklishin:
@acogoluegnes FYI, I discussed this with @captainju on Slack and suggested that we would consider such a contribution.

src/main/java/com/rabbitmq/client/ConnectionFactory.java
captainju: I used the same charset as in uriDecode line 378, which is used to decode username password and vhost.
```
    private static String uriDecode(String s) {
        try {
            // URLDecode decodes '+' to a space, as for
            // form encoding.  So protect plus signs.
            return URLDecoder.decode(s.replace("+", "%2B"), "US-ASCII");
        }
        catch (IOException e) {
            throw new RuntimeException(e);
        }
    }
```
https://github.com/rabbitmq/rabbitmq-java-client/pull/672/files#diff-045e321e0fb12f11085653a7ed09f16518f9f21a7c80990c30720958dc17e954R378

acogoluegnes:
Thanks!

PR #674, opened 2021-03-22
Add support for reading unsigned short "u".

gustf:
## Proposed Changes

Add support for reading unsigned short "u".

https://www.rabbitmq.com/amqp-0-9-1-errata.html#section_3

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/ValueReader.java
+++ b/src/main/java/com/rabbitmq/client/impl/ValueReader.java
@@ -194,6 +194,9 @@ public class ValueReader
           case 's':
               value = in.readShort();
               break;
+          case 'u':
+              value = in.readUnsignedShort();
+              break;
           case 't':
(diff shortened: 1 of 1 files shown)


michaelklishin (approved):

michaelklishin:
I'm curious what client you use on the publishing end and why you need this relatively rarely used data type but fair enough :)

gustf:
Hi @michaelklishin 

I am using AMQPStorm, https://www.amqpstorm.io
Which in turn are using pamqp, https://pamqp.readthedocs.io/

Writing it here
https://github.com/gmr/pamqp/blob/b269d6f7f3083504931079c6d1f1113ded830f02/pamqp/encode.py#L315


Another thing, just noticed that there is no support for unsigned int either, ok to add a PR for that one also?

michaelklishin:
@gustf sure.

Issue #689, opened 2021-06-18
Allow maintaining of address order with auto recovery


ValSmith:
According to the documentation you can use an AddressResolver to choose a desired host first. I created a resolver that returns [host1,host2,host3] but then found that the connections were all to host2 even though host 1 was available. After a while I discovered that when autoRecovery=true the address list is always shuffled. Could there be a possibility to maintain the list order when using auto recovery? Or is there a very intentional reason this is not allowed?

michaelklishin:
This is by design. 

Depending on the order of endpoints doesn't sound like a good idea. Any client can [connect to any cluster node](https://www.rabbitmq.com/clustering.html#clustering-and-clients). If a "preferred" node is unreachable, attempts to connect to it sequentially will simply increase the amount of time it takes to reach to a reachable endpoint with a "lower priority".

PR #690, opened 2021-06-18
Added option to maintain AddressResolver address order for AutoRecove…

ValSmith:
Allows the option to maintain the address order from the address resolver when using an auto recoverable connection.  The default behavior is unchanged. This pull request is for the issue https://github.com/rabbitmq/rabbitmq-java-client/issues/689

### DIFF
--- a/src/main/java/com/rabbitmq/client/ConnectionFactory.java
+++ b/src/main/java/com/rabbitmq/client/ConnectionFactory.java
@@ -126,6 +126,7 @@ public class ConnectionFactory implements Cloneable {
     private boolean automaticRecovery               = true;
     private boolean topologyRecovery                = true;
     private ExecutorService topologyRecoveryExecutor;
+    private boolean maintainAddressOrder;
     
     // long is used to make sure the users can use both ints
     // and longs safely. It is unlikely that anybody'd need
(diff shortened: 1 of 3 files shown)


michaelklishin:
Thank you for taking the time to put together a PR. As I explained in #689, this behavior is by design. Any client can connect to any node and "endpoint priority" is not a concept I feel we need since it doesn't actually exist in RabbitMQ and any client I can think of. 

`ConnectionFactory` already has so many options that we would prefer to see a lot of demand for a feature before introducing it.

michaelklishin:
What we could potentially do is to move the shuffling to the currently available resolvers and let someone introduce their own that maintains a sequential order. I don't know how many people with custom resolves there are, though, and how many of them would prefer to avoid the reshuffling or care about it. But at least most users do not use custom resolvers.

michaelklishin:
I later thought of another option: we could introduce a base class that does reshuffling in a "post-resolution" step and base all existing implementations  on it. Let me have a quick go at it.

ValSmith:
That could be a good idea.  There could be a default implementation of AddressResolver that does the shuffling then the connection factories don't have to think about it at all.  You could initialize with a static AddressResolver.build([addresses]).  Our specific use case is that we have a rabbit cluster running in various availability zones in AWS and an application that is also spread out so we want each application node to connect to the closer rabbit node if possible.

michaelklishin:
Yup, a default implementation in `AddressResolver` is what I ended up with, see #691.

PR #691, opened 2021-06-20
Move resolved address shuffling to AddressResolver

michaelklishin:
so that it can be overridden by implementations, e.g. to perform
no shuffling at all.

References #690.


michaelklishin:
@acogoluegnes WDYT?

acogoluegnes (approved):

PR #693, opened 2021-07-02
Allow changing queue names during recovery

vikinghawk:
## Proposed Changes

Allow a consumer to register a callback to change the name of a queue during recovery. This is useful when dealing with transient and auto-delete queues that are susceptible to 404 NOT_FOUND client recovery errors and binding corruption issues during cluster node cycles/upgrades. When a client reconnects and redeclares the queue, we expose a timing issue due to the server trying to cleanup the queue at the same time the client is coming back to redeclare it.

The easy existing alternative is to use server named queues. However, with that consumers have no ability to provide a useful queue name that makes it easy to tell what application a queue belongs too from a monitoring/troubleshooting perspective.

This enhancement allows for applications to still name queues in meaningful ways, but to then support modifying that name during a recovery to avoid the timing issues described above.

## Types of Changes

## Checklist

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
+++ b/src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
@@ -96,6 +96,8 @@ public class AutorecoveringConnection implements RecoverableConnection, NetworkC
 	private final Predicate<ShutdownSignalException> connectionRecoveryTriggeringCondition;
 
 	private final RetryHandler retryHandler;
+	
+	private final RecoveredQueueNameSupplier recoveredQueueNameSupplier;
 
     public AutorecoveringConnection(ConnectionParams params, FrameHandlerFactory f, List<Address> addrs) {
         this(params, f, new ListAddressResolver(addrs));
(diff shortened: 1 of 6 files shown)


src/main/java/com/rabbitmq/client/impl/recovery/AutorecoveringConnection.java
```diff
@@ -782,11 +790,7 @@ public void recoverQueue(final String oldName, RecordedQueue q, boolean retry) {
                         this.propagateQueueNameChangeToConsumers(oldName, newName);
                         // bug26552:
                         // remove old name after we've updated the bindings and consumers,
-                        // plus only for server-named queues, both to make sure we don't lose
-                        // anything to recover. MK.
-                        if(q.isServerNamed()) {
-                            deleteRecordedQueue(oldName);
-                        }
+                        deleteRecordedQueue(oldName);
```
vikinghawk, line 793: wanted to call out that this is removing the q.isServerNamed() check here. But i don't think that is an issue as i don't believe their would have historically been anyway for the name to change unless it was server named.

michaelklishin (approved):

michaelklishin:
Thank you!

michaelklishin:
@vikinghawk can you please submit this PR against `main`?

michaelklishin:
Actually no need to do that, I managed to merge `5.x.x-stable` into `main`.

Issue #700, opened 2021-07-28
Large message deliveries with TLS and NIO enabled results in a "buffer closed" exception in SslEngineHelper


YuryAndr:
Hello!

There was an issue https://github.com/rabbitmq/rabbitmq-java-client/issues/307 .

And it looks like it reproduced again.

**Versions:**
Server - RabbitMQ 3.8.11

Client -
```
<dependency>
    <groupId>com.rabbitmq</groupId>
    <artifactId>amqp-client</artifactId>
    <version>5.13.0</version>
</dependency>
```

**Steps to reproduce**
- SSL and NIO are enabled
- Send a large message to queue
- Start consumer of the queue

**Result**
Connection is closed with error
```
2021/07/28 19:50:17.240 ERROR [rabbitmq-connection-shutdown-amqp://guest@127.0.0.1:5671/test] c.r.c.i.ForgivingExceptionHandler - An unexpected connection driver error occurred 
javax.net.ssl.SSLException: Unrecognized record version (D)TLS-0.0 , plaintext connection?
	at java.base/sun.security.ssl.SSLEngineInputRecord.bytesInCompletePacket(SSLEngineInputRecord.java:98)
	at java.base/sun.security.ssl.SSLEngineInputRecord.bytesInCompletePacket(SSLEngineInputRecord.java:64)
	at java.base/sun.security.ssl.SSLEngineImpl.readRecord(SSLEngineImpl.java:548)
	at java.base/sun.security.ssl.SSLEngineImpl.unwrap(SSLEngineImpl.java:443)
	at java.base/sun.security.ssl.SSLEngineImpl.unwrap(SSLEngineImpl.java:422)
	at java.base/javax.net.ssl.SSLEngine.unwrap(SSLEngine.java:634)
	at com.rabbitmq.client.impl.nio.SslEngineFrameBuilder.somethingToRead(SslEngineFrameBuilder.java:49)
	at com.rabbitmq.client.impl.nio.FrameBuilder.readFrame(FrameBuilder.java:71)
	at com.rabbitmq.client.impl.nio.NioLoop.run(NioLoop.java:156)
	at java.base/java.lang.Thread.run(Thread.java:834)
2021/07/28 19:50:17.241 ERROR [rabbitmq-connection-shutdown-amqp://guest@127.0.0.1:5671/test] c.r.c.i.ForgivingExceptionHandler - An unexpected connection driver error occurred 
javax.net.ssl.SSLException: Buffer closed
[...]

PR #701, opened 2021-07-29
Support underflow handling without thread sleep

YuryAndr:
## Proposed Changes

Purpose of PR is to fix bug -https://github.com/rabbitmq/rabbitmq-java-client/issues/700

According to https://docs.oracle.com/javase/7/docs/technotes/guides/security/jsse/JSSERefGuide.html we should perform read next bytes of message when `SSLEngineResult` indicates `BUFFER_UNDERFLOW` state.

In versions before `5.*.*`  it was fixed using retries with Thread.sleep() https://github.com/rabbitmq/rabbitmq-java-client/blob/87aae532705490fa429afe34326a0ebbdd83e1d3/src/main/java/com/rabbitmq/client/impl/nio/NioHelper.java

But versions `5.*.*` don't include this fix. Also It is not good idea to use Thread.sleep() on nio thred.

Current PR provides fetching chunks of bytes using standard NIO flow. `SocketChannelFrameHandlerState` provides `isUnderflowHandlingEnabled` attribute to track current state.

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/nio/SocketChannelFrameHandlerState.java
+++ b/src/main/java/com/rabbitmq/client/impl/nio/SocketChannelFrameHandlerState.java
@@ -26,6 +26,8 @@ import java.io.IOException;
 import java.nio.ByteBuffer;
 import java.nio.channels.SelectionKey;
 import java.nio.channels.SocketChannel;
+import java.util.concurrent.atomic.AtomicBoolean;
+
 
 /**
  *
@@ -70,6 +72,8 @@ public class SocketChannelFrameHandlerState {
 
     final FrameBuilder frameBuilder;
 
+    private final AtomicBoolean isUnderflowHandlingEnabled = new AtomicBoolean(false);
+
     public SocketChannelFrameHandlerState(SocketChannel channel, NioLoopContext nioLoopsState, NioParams nioParams, SSLEngine sslEngine) {
         this.channel = channel;
         this.readSelectorState = nioLoopsState.readSelectorState;
@@ -105,7 +109,7 @@ public class SocketChannelFrameHandlerState {
             this.outputStream = new DataOutputStream(
                 new SslEngineByteBufferOutputStream(sslEngine, plainOut, cipherOut, channel)
             );
-            this.frameBuilder = new SslEngineFrameBuilder(sslEngine, plainIn, cipherIn, channel);
+            this.frameBuilder = new SslEngineFrameBuilder(sslEngine, plainIn, cipherIn, channel, isUnderflowHandlingEnabled);
         }
 
     }
@@ -176,11 +180,14 @@ public class SocketChannelFrameHandlerState {
 
     void prepareForReadSequence() throws IOException {
         if(ssl) {
-            cipherIn.clear();
-            plainIn.clear();
+            if (!isUnderflowHandlingEnabled.get()) {
+                cipherIn.clear();
+                cipherIn.flip();
+            }
 
-            cipherIn.flip();
+            plainIn.clear();
             plainIn.flip();
+
         } else {
             NioHelper.read(channel, plainIn);
             plainIn.flip();
@@ -189,6 +196,15 @@ public class SocketChannelFrameHandlerState {
 
     boolean continueReading() throws IOException {
         if(ssl) {
+            if (isUnderflowHandlingEnabled.get()) {
+                int bytesRead = NioHelper.read(channel, cipherIn);
+                if (bytesRead == 0) {
+                    return false;
+                } else {
+                    cipherIn.flip();
+                    return true;
+                }
+            }
             if (!plainIn.hasRemaining() && !cipherIn.hasRemaining()) {
                 // need to try to read something
                 cipherIn.clear();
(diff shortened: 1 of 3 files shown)


acogoluegnes:
You beat me to it, I just pushed #702. You got the problem right, but I think #702 keeps the changes more local, WDYT?

YuryAndr:
This PR takes into account `cipherIn.clear()` before next chunk reading at SocketChannelFrameHandlerStat. So there is no losses of previous chunks.

src/main/java/com/rabbitmq/client/impl/nio/SocketChannelFrameHandlerState.java
```diff
@@ -176,11 +180,14 @@ void endWriteSequence() {
...
-            cipherIn.clear();
-            plainIn.clear();
+            if (!isUnderflowHandlingEnabled.get()) {
+                cipherIn.clear();
+                cipherIn.flip();
+            }
 
-            cipherIn.flip();
+            plainIn.clear();
             plainIn.flip();
```
acogoluegnes, line 189: Can you elaborate on the need of these operations?

src/main/java/com/rabbitmq/client/impl/nio/SslEngineFrameBuilder.java
```diff
@@ -59,18 +66,18 @@ protected boolean somethingToRead() throws IOException {
                         throw new SSLException("buffer overflow in read");
                     case BUFFER_UNDERFLOW:
                         cipherBuffer.compact();
-                        int read = NioHelper.read(channel, cipherBuffer);
-                        if (read == 0) {
-                            return false;
-                        }
-                        cipherBuffer.flip();
-                        break;
+                        underflowHandling = true;
```
acogoluegnes, line 69: OK, removing some reading logic and putting it in the class that was already doing, nice.

acogoluegnes (changes requested):
Nice contribution, thanks! Do you think it would be possible to get rid of the `isUnderflowHandlingEnabled` AtomicBoolean, keep the information in the frame builder, and expose with a package-protected method? The state could reset the flag with another package-protected method. This would avoid to use an atomic class (we don't need the guarantees it provides as this is all mono-threaded).

src/main/java/com/rabbitmq/client/impl/nio/SocketChannelFrameHandlerState.java
YuryAndr: `cipherIn` should aggregate chunks of bytes on each read. So we should not clear it before each chunk reading.

YuryAndr:
>Do you think it would be possible to get rid of the isUnderflowHandlingEnabled AtomicBoolean

isUnderflowHandlingEnabled was converted to private attribute of SslEngineFrameBuilder

src/main/java/com/rabbitmq/client/impl/nio/SocketChannelFrameHandlerState.java
YuryAndr: Also in my implementation `continueReading` returns false when there is no data for next chunk. So there is no infinite loop around `while(continueReading())`. NIO thread can check selector state before trying to get bytes.

acogoluegnes (approved):

michaelklishin:
@YuryAndr thank you for contributing. We will back port after testing but you are welcome to build this client from source and try it out on your actual workload one more time ;)

acogoluegnes:
@YuryAndr [5.13.1-SNAPSHOT](https://oss.sonatype.org/content/repositories/snapshots/com/rabbitmq/amqp-client/5.13.1-SNAPSHOT/) is available if you want to give it a try. I'll release 5.13.1 if it works.

YuryAndr:
@acogoluegnes Thank you! It was verified against our setup. 

Also it was found that PR had fixed another related issue. 

It was noticed that error in BUFFER_UNDERFLOW state handling leads to out of memory error.
```
Exception in thread "rabbitmq-nio" java.lang.OutOfMemoryError: Java heap space
	at com.rabbitmq.client.impl.nio.FrameBuilder.readFrame(FrameBuilder.java:96)
	at com.rabbitmq.client.impl.nio.NioLoop.run(NioLoop.java:156)
	at java.base/java.lang.Thread.run(Thread.java:829)
```

Because FrameBuilder tries to read random data from buffer and tries to create huge byte array about 1.7 GB.

image

It was also verified that there are no such memory issues on **5.13.1-SNAPSHOT**.

YuryAndr:
@acogoluegnes Could you please also provide patch for `5.7.*` version? Because we have modules with dependency on Spring `5.2.*` version and spring-amqp `2.2.*` https://github.com/spring-projects/spring-amqp/blob/2.2.x/build.gradle  And it depends on `5.7.3`. version of java client.

michaelklishin:
We generally do not release patches for older minor series, in particular six series behind. @YuryAndr if you are willing to do all the backporting and testing work for `5.7.x`, we can cut a new release.

michaelklishin:
So yeah, in addition to backporting to `5.13.x-stable` we'd have to backport to `5.7.x-stable` and then produce a release.
I don't know if we have the pipelines to produce `5.7.x` releases but I guess we can generate a one-off.

YuryAndr:
Next PR provides cherry picked commits for `5.7.x `https://github.com/rabbitmq/rabbitmq-java-client/pull/703

acogoluegnes:
OK, I'll see what I can to release a 5.7.x, but it won't be before several weeks.

PR #716, opened 2021-11-05
Fix handshake with NIO on TLS 1.3

### LINKED ISSUE #715: TLS handshake hangs with NIO and TLS 1.3
Original mailing list post: https://groups.google.com/g/rabbitmq-users/c/LGlpHNx5QIQ/m/AobNKVegAAAJ
Related discussion: https://github.com/rabbitmq/rabbitmq-java-client/discussions/712.

acogoluegnes:
The unwrapping does not work the same way between TLS 1.2 and 1.3.
This commit makes the unwrapping more reliable by getting
the number of bytes consumed in the unwrapping and then set the
position of the reading ByteBuffer accordingly to the number of bytes.
With TLS 1.3, the unwrapping seems to read the whole content of
the buffer and to extract only the first record, so the rewinding
is necessary.

The commit also adds some debug logging, adds tests on TLS 1.2 and 1.3,
and re-arranges the TLS test (add utility class).

Fixes #715

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/nio/SslEngineHelper.java
+++ b/src/main/java/com/rabbitmq/client/impl/nio/SslEngineHelper.java
@@ -1,4 +1,4 @@
-// Copyright (c) 2007-2020 VMware, Inc. or its affiliates.  All rights reserved.
+// Copyright (c) 2007-2021 VMware, Inc. or its affiliates.  All rights reserved.
 //
 // This software, the RabbitMQ Java client library, is triple-licensed under the
 // Mozilla Public License 2.0 ("MPL"), the GNU General Public License version 2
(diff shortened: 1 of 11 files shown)


michaelklishin:
@yaytay can you please help us verify this PR?

Yaytay:
Happy to, but it might be early next week before I get the chance.

Yaytay:
Works for me.
I built it on a Debian VM, copied to my Windows host and specified it explicitly in the pom using systemPath.
Then the code as sent before, using TLv1.3 works.
Thanks.

acogoluegnes:
@acogoluegnes [5.14.0.RC1](https://groups.google.com/g/rabbitmq-users/c/B3XDSwHFN28/m/m0l06i3NBgAJ) is available.

PR #878, opened 2022-11-14
Remove dependency on java.sql.Timestamp

acogoluegnes:
Only one condition test refers to java.sql.Timestamp and it's unnecessary anyway because it also refers to java.util.Date (Timestamp inherits from Date). Only the Date API is then used to marshall and unmarshall values.

This avoids a dependency on the java.sql module, which some SDK does not seem to contain (e.g. JavaFX, reported by a user). This should also avoid needing this module in some Docker images (e.g. PerfTest, which pulls this module in its stripped JRE just for this reference apparently).

Mailing list thread: https://groups.google.com/g/rabbitmq-users/c/ORX91PwYhnw/m/xlMR9JTTAQAJ


michaelklishin (approved):
Nice catch!

PR #1017, opened 2023-05-02
Support Micrometer observation

acogoluegnes:


src/main/java/com/rabbitmq/client/observation/micrometer/DefaultConsumeObservationConvention.java
```diff
@@ -0,0 +1,68 @@
...
+    return KeyValues.of(
+        LowCardinalityTags.MESSAGING_OPERATION.withValue("consume"),
+        LowCardinalityTags.MESSAGING_SYSTEM.withValue("rabbitmq"));
+  }
+
+  @Override
+  public KeyValues getHighCardinalityKeyValues(ConsumeContext context) {
+    return KeyValues.of(
+        HighCardinalityTags.MESSAGING_ROUTING_KEY.withValue(context.getRoutingKey()),
+        HighCardinalityTags.MESSAGING_DESTINATION_NAME.withValue(context.getExchange()));
```
acogoluegnes, line 66: In the case of `consume`, should the destination be the queue the consumer is registered to or the exchange (like it is now)? In this context the exchange is rather the "origin" or the "source" and the queue is the place where the message ended up in. I'm not sure what kind of search users are expected to do but I'd say they would like to differentiate between exchanges and queues. @marcingrzejszczak WDYT?

ikvmw:
Please note this OT integration - https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261. We should aim for same semantic and tags.

stebet:
> 

Awesome! Don't hesitate to ping me in the .NET PR as the spec has probably changed a few times (although I've tried to keep up) since I started this work, and let's keep things in sync if you feel the .NET version should name things differently or vice/versa. Would also be great to get some feedback on tags and operation names from @lmolkova @trask and hopefully more people with OTel knowledge if they have the time.

ikvmw:
Hi, I created an md file to synchronize and track - https://github.com/rabbitmq/contribute/blob/main/observability/clients.md. Please take a look and change/add missing parts

trask:
> > Please note this OT integration - [rabbitmq/rabbitmq-dotnet-client#1261](https://github.com/rabbitmq/rabbitmq-dotnet-client/pull/1261). We should aim for same semantic and tags.
> 
> Awesome! Don't hesitate to ping me in the .NET PR as the spec has probably changed a few times (although I've tried to keep up) since I started this work, and let's keep things in sync if you feel the .NET version should name things differently or vice/versa. Would also be great to get some feedback on tags and operation names from @lmolkova @trask and hopefully more people with OTel knowledge if they have the time.

hi! @lmolkova and I would of course recommend instrumenting with OpenTelemetry API 😄

acogoluegnes:
@stebet Thanks for chiming in. I had a look at the .NET client PR. It's similar in concepts to this one: we are using an abstraction (ActivitySource / Micrometer) and we are using OTel naming conventions. I just want to check something with you: is the `receive` activity "around" the `process` activity? It seems to me it is and that they are not one after the other. OTel is not super clear on this part.

acogoluegnes:
@stebet I had a look at the [OTel Java instrumentation agent](https://github.com/open-telemetry/opentelemetry-java-instrumentation/tree/main/instrumentation/rabbitmq-2.7/javaagent/src/main/java/io/opentelemetry/javaagent/instrumentation/rabbitmq) and it uses `receive` for `basic.get` and `process` for asynchronous deliveries (from `basic.consume`).

trask:
> It's similar in concepts to this one: we are using an abstraction (ActivitySource / Micrometer)

the OpenTelemetry and .NET community aligned and ActivitySource is (part of) the OpenTelemetry API for .NET: https://opentelemetry.io/docs/instrumentation/net/manual/#add-tags-to-an-activity

(while Micrometer Tracing is Spring-specific)

lmolkova:
I might miss something, but my understanding was that micrometer-tracing is an abstraction layer over otel or brave. With brave slowly becoming legacy, what's the benefit of maintaining an extra layer of abstraction?

lmolkova:
> @stebet Thanks for chiming in. I had a look at the .NET client PR. It's similar in concepts to this one: we are using an abstraction (ActivitySource / Micrometer) and we are using OTel naming conventions. I just want to check something with you: is the `receive` activity "around" the `process` activity? It seems to me it is and that they are not one after the other. OTel is not super clear on this part.

The `receive` operation over `process` is a trick to correlate them - receive operation does not necessarily know what it's going to be received.
OTel messaging SIG is working on a better version of the spec (https://github.com/open-telemetry/oteps/pull/220) - here's we'll have two receive models:
- receive (poll)  - it will be a single span. It should have a link to the received message. It's not yet possible to add link after span starts, so it should be started later with a fake start time.
- deliver (consume) - it mostly matches the current `process` and can be a child of context in the message.

It does not contradict current spec (which is indeed quite vague), so can be done right away. And I hope the OTEP above gives more clarity on implementation

marcingrzejszczak:
> I might miss something, but my understanding was that micrometer-tracing is an abstraction layer over otel or brave.

Yes it is

> With brave slowly becoming legacy

Where this assumption comes from?

> what's the benefit of having an extra layer of abstraction?

That we control the api and can replace one tracer into another. Micrometer is very serious about not breaking backward compatibility and the history of current tracing projects such as open tracing or open census warns about what can happen when you rely directly on an api.

Of course you can use micrometer observation with otels api directly - you don't have to use micrometer tracing

src/test/java/com/rabbitmq/client/test/functional/MicrometerObservationCollectorMetrics.java
```diff
@@ -111,6 +120,17 @@ private abstract static class IntegrationTest extends SampleTestRunner {
     public TracingSetup[] getTracingSetup() {
       return new TracingSetup[] {TracingSetup.IN_MEMORY_BRAVE, TracingSetup.ZIPKIN_BRAVE};
     }
+
+    void runWithNullingObservation(ObservationRegistry registry, Tracer tracer, Observation.CheckedRunnable<?> runnable) {
+      Observation noParentObservation = Observation.createNotStarted("null_observation", registry);
```
marcingrzejszczak, line 125: With Micrometer 1.10.8 I think we can change this whole method to become `new NullObservation(registry).observeChecked(runnable)`

src/main/java/com/rabbitmq/client/observation/micrometer/MicrometerObservationCollector.java
```diff
@@ -0,0 +1,212 @@
...
+                }
+                DeliverContext context =
+                    new DeliverContext(
+                        response.getEnvelope().getExchange(),
+                        response.getEnvelope().getRoutingKey(),
+                        queue,
+                        headers,
+                        response.getBody() == null ? 0 : response.getBody().length);
+                Observation observation =
+                    RabbitMqObservationDocumentation.RECEIVE_OBSERVATION.observation(
```
artembilan, line 129: Is its really intentional to start another observation over here?
You already got that `Observation.createNotStarted("rabbitmq.receive")` and its `observe()`.
Why do we need this child one, please?

artembilan (changes requested):

src/main/java/com/rabbitmq/client/observation/micrometer/MicrometerObservationCollector.java
marcingrzejszczak: So we have 2 observations here.

- `rabbitmq.receive` is the observation that is responsible for polling. Some trace (e.g. with id X) might have been initiated earlier and suddenly someone calls `basicGet` the `rabbitmq.receive` observation will warp an operation of polling a message from the broker
- `RECEIVE_OBSERVATION` is the `ReceiverContext` based operation that will continue a trace that got propagated through the messaging headers. That means that in the RabbitMq message there could be another trace (e.g. with id Y)

Let's look at it in the following way:

- Application A initiates trace with id X -> calls some logic -> polls for the message from RabbitMq (rabbitmq.receive) <THE TRACE X FINISHES HERE>
- Application B initiates trace with id Y -> calls some logic -> sends a message to RabbitMQ with tracing header containing Y -> Application A takes the message from the queue and then continues trace id Y (RECEIVE_OBSERVATION)

Since `basicGet` is a synchronous operation we were thinking of giving the user a switch to leave the `RECEIVE_OBSERVATION` started, and a scope would be opened and then it's up to the user to stop it and stop the scope. That way they will be able to continue the trace id `Y`

artembilan: But you do that manually here just like this a bit bellow:
```
                observation.start();
                observation.stop();
```
So, yeah, you connect this `RECEIVE_OBSERVATION` to the propagated one, but it is closed immediately. 
Whatever end-user would do with the propagated trace over header afterward would be just another, parallel child span, not connected to this one. Plus looks this one is going to be a child of that `rabbitmq.receive` in the beginning of this method. No? Is the parent overridden from the value from headers?

I may guess that it is OK to have a simple span connected to that propagated trace just to fulfill the the whole picture of services interaction, but is it really an intention? Wouldn't it be better to leave such a propagation up to end-user with respective API?

marcingrzejszczak: > So, yeah, you connect this RECEIVE_OBSERVATION to the propagated one, but it is closed immediately.
Whatever end-user would do with the propagated trace over header afterward would be just another, parallel child span, not connected to this one. 

As I said this code 

```java
 observation.start();
                observation.stop();
```

would have to change to sth like

```java
observation.start();
if (someOptionIsTurnedOn) {
  observation.openScope();
} else {
  observation.stop();
}
```

and then the user will have to do sth like this after some listener code gets executed

```java

Scope scope = observationRegistry.getCurrentObservationScope();
if (scope != null) {
  scope.close();
  scope.getCurrentObservation().stop();
}
```

> Plus looks this one is going to be a child of that rabbitmq.receive in the beginning of this method. No? Is the parent overridden from the value from headers?

Yes, the child will be overridden. So if you started trace X then due to the fact that the message contains X headers, the trace id will continue.

> I may guess that it is OK to have a simple span connected to that propagated trace just to fulfill the the whole picture of services interaction, but is it really an intention? Wouldn't it be better to leave such a propagation up to end-user with respective API?

I don't follow, what do you mean exactly?

Issue #330, opened 2017-11-09
ConnectionFactory: fluent style


rrrrrrrrramba:
It would be useful if the API of `ConnectionFactory` supported fluent style. For example:

    Connection c = new ConnectionFactory().setUri(...).newConnection();
    Connection c = new ConnectionFactory().load(...).newConnection();

Most of the times, this involves returning `this` in the setters.

I know this might be a personal opinion, but what do you think?

acogoluegnes:
Good idea. `NioParams`, introduced in 4.0, already uses a fluent API. This is a breaking change though, so it must be introduced in the next major release, that is 6.0.

rrrrrrrrramba:
Great, thanks!

enniokerber:
Hey @acogoluegnes,

I feel like this might be a topic suited for an Open-Source beginner. Would you mind assigning this to me? I'd be eager to work on it.

Greetings,
Ennio

michaelklishin:
Thanks for contributing, @enniokerber. @acogoluegnes perhaps we can ship a small (source-level) backwards compatible like #1038 in `5.18`?

acogoluegnes:
#1038 is not exactly a backward-compatible change: applications or libraries using reflection could depend on the method return type. This is very unlikely though, so I'd say we can backport this to 5.x. Is that OK with you @michaelklishin?

PR #1038, opened 2023-05-20
Configure ConnectionFactory instances using fluent API pattern

enniokerber:
## Proposed Changes

This pull request adds the ability to configure ConnectionFactory instances using the fluent API pattern as proposed in #330.
The aim of this is to make configuration of new connections easier and require less code.

**Before** (extracted from SslContextFactoryTest)

```        
        doTestSetSslContextFactory(() -> {
            ConnectionFactory connectionFactory = new ConnectionFactory();
            connectionFactory.useBlockingIo();
            connectionFactory.setAutomaticRecoveryEnabled(true);
            return connectionFactory;
        });
```

**After**

```
        doTestSetSslContextFactory(() -> new ConnectionFactory()
                .useBlockingIo()
                .setAutomaticRecoveryEnabled(true)
        );
```

As you can see, using this pattern increases readability and reduces boilerplate code.

## Types of Changes

As proposed in #330 already, I also encourange deploying this with the new major version 6 as the API of ConnectionFactory changes due to this feature.

## Checklist

### DIFF
--- a/src/main/java/com/rabbitmq/client/ConnectionFactory.java
+++ b/src/main/java/com/rabbitmq/client/ConnectionFactory.java
@@ -25,8 +25,6 @@ import com.rabbitmq.client.impl.recovery.TopologyRecoveryFilter;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-import java.util.Map.Entry;
-import java.util.function.BiConsumer;
 import javax.net.SocketFactory;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLSocketFactory;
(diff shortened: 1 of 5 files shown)


michaelklishin:
I don't think it really "reduces boilerplate code" in your example, it's basically one fewer line, and usually with the Builder pattern you have an explicit termination method (`.build()` or similar).

You are welcome to submit a PR that updates `ConnectionFactory` to make that possible. It will be considered for the next minor release.

michaelklishin:
That said, thanks for taking the time to contribute. I don't have any objections to having
an API like this, assuming the existing code does not have to be modified (recompiling is OK for the next minor release).

michaelklishin:
To confirm: some instantiations of `ConnectionFactory` in tests still use the original API,
therefore proving (source-level) backwards compatibility.

michaelklishin (approved):
Thank you!

Issue #434, opened 2018-12-18
Channel#abort discards exceptions which it declares as thrown


thiagomiranda3:
On the code below that I got from master:
`/** Public API - {@inheritDoc} */
    @Override
    public void abort(int closeCode, String closeMessage)
        throws IOException
    {
        try {
          close(closeCode, closeMessage, true, null, true);
        } catch (IOException _e) {
        /* ignored */
        } catch (TimeoutException _e) {
          /* ignored */
        }
    }`

The method abort throw an IOException that I need to catch when calling this method. But this exception is silently discarded inside them and in the documentation is written that too. So I was suspicious that the documentation was not up to date and had to check the source to verify their validity. Wouldn't be possible to remove this throw? I am not sure with it is necessary in another classes that override this method.

acogoluegnes:
This makes sense, considering the semantics mentioned in the API documentation and the consistency with `Connection#abort`. This is a breaking change though (code catching the `IOException` won't compile anymore), so it must go into 6.0.

PR #1039, opened 2023-05-21
feat(Channel): Remove checked exception from interface Channel.abort(...) as it should silently discard any exceptions

enniokerber:
## Proposed Changes
As proposed in #434, I eliminated the "throws" clause from the Channel.abort(...) interface in order to align with the desired behaviour, which is also document above the method definitions: 
_Any encountered exceptions in the close operation are silently discarded._

At the current time, consumers calling the abort(...) method will have to manually check for IOException by using a try-catch-block, which is obsolete and just unnecessarily blows up code.

## Types of Changes

As already mentioned inside #434: 
_This is a breaking change though (code catching the IOException won't compile anymore), so it must go into 6.0._

## Checklist

## Further Comments

It seems as there are no existing tests for class `AutorecoveringChannel`. This might be something that could be improved in the future. I just added some initial tests for the abort methods.

### DIFF
--- a/src/main/java/com/rabbitmq/client/Channel.java
+++ b/src/main/java/com/rabbitmq/client/Channel.java
@@ -89,7 +89,7 @@ public interface Channel extends ShutdownNotifier, AutoCloseable {
      * Forces the channel to close and waits for the close operation to complete.
      * Any encountered exceptions in the close operation are silently discarded.
      */
-    void abort() throws IOException;
+    void abort();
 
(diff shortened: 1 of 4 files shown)


michaelklishin:
Autorecovering channels are tested, just like connections, in connection recovery integration tests.

acogoluegnes (approved):

acogoluegnes:
Thanks!

PR #1057, opened 2023-06-08
Do not confirmSelect more than once per channel

### LINKED ISSUE #1056: Do not confirmSelect more than once per channel
### Is your feature request related to a problem? Please describe.

Reactor's RabbitMQ [`sendWithTypedPublishConfirms()`](https://github.com/reactor/reactor-rabbitmq/blob/5dde6ee8f51de48945ceddf5e533b6455503b45a/src/main/java/reactor/rabbitmq/Sender.java#L278-L290) always invokes the `confirmSelect()` method of the underlying channel which causes a blocking RPC call. 

This is especially detrimental to performance when sending a small amount of messages per Publisher/Flux, as the channel is put into confirm mode even if it already was (when using a channel pool).

Following [RabbitMQ's best practices](https://www.rabbitmq.com/tutorials/tutorial-seven-java.html#:~:text=Confirms%20should%20be%20enabled%20just%20once%2C%20not%20for%20every%20message%20published.), confirms should be enabled once per channel. In fact, there's no API to disable confirms once enabled on a channel.

### Describe the solution you'd like

`ChannelN` should keep track if `confirmSelect()` was successfully enabled, and if so, future invocations of this method return a `Confirm.SelectOk` without making a RPC call.

I can contribute a PR if there's interest in this solution 🙂 

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

sergio91pt:
## Proposed Changes

In order to avoid unnecessary blocking RPC calls and conform to best practices, the Channel now checks if it already activated confirm mode before sending a confirm.select RPC call.

If confirm mode is already activated, calling confirmSelect() again returns immediately without sending an RPC call.

Closes #1056

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #NNNN)

## Checklist

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/ChannelN.java
+++ b/src/main/java/com/rabbitmq/client/impl/ChannelN.java
@@ -82,6 +82,9 @@ public class ChannelN extends AMQChannel implements com.rabbitmq.client.Channel 
     private final SortedSet<Long> unconfirmedSet =
             Collections.synchronizedSortedSet(new TreeSet<Long>());
 
+    /** Whether the confirm select method has been successfully activated */
+    private boolean confirmSelectActivated = false;
+
     /** Whether any nacks have been received since the last waitForConfirms(). */
     private volatile boolean onlyAcksReceived = true;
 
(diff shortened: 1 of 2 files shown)


michaelklishin:
In a design briefly discussed on our internal Slack with @acogoluegnes, we concluded that it would be idea to use an `AtomicBoolean`. On the other hand, channels are not mean to be shared between threads as much as possible, and *never* for publishers. So this PR looks OK as it is.

acogoluegnes:
It's an optimization for a network that is already idempotent and channels are not meant to be used by several threads, so we can save a bit of memory by using a boolean. We could also make the property volatile, this should have an impact only on calls to `confirmSelect()`. I'd lean towards the `volatile boolean`, but I'm not super opinionated about it.

acogoluegnes:
@sergio91pt Can you increase the timeout in the test, to 10 seconds? It passes locally, but not on CI.

sergio91pt:
@acogoluegnes Can you run the pipeline again? I refactored the test and it should pass now. I already run it 3000 times through the IDE and it doesn't seem flaky anymore.
It now uses a thread, since I noticed that convenient Executor was actually being passed to the code under test, so I shouldn't be using it 😄 

Regarding the boolean, I also was considering a volatile boolean initially but the usage of `nextPublishSeqNo` made it clear this method is not intended to be thread safe, so I ended leaving it as a normal boolean.

michaelklishin (approved):

PR #1063, opened 2023-06-15
Add max inbound message size to ConnectionFactory

### LINKED ISSUE #1062: Add ability to specify maximum message size

acogoluegnes:
The default value is 64 MiB.

Fixes #1062

### DIFF
--- a/src/main/java/com/rabbitmq/client/impl/CommandAssembler.java
+++ b/src/main/java/com/rabbitmq/client/impl/CommandAssembler.java
@@ -107,7 +107,9 @@ final class CommandAssembler {
             long bodySize = this.contentHeader.getBodySize();
             if (bodySize >= this.maxBodyLength) {
                 throw new IllegalStateException(format(
-                    "Message body is too large (%d), maximum size is %d",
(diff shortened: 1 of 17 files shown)


src/main/java/com/rabbitmq/client/impl/CommandAssembler.java
```diff
@@ -99,7 +104,14 @@ private void consumeMethodFrame(Frame f) throws IOException {
     private void consumeHeaderFrame(Frame f) throws IOException {
         if (f.getType() == AMQP.FRAME_HEADER) {
             this.contentHeader = AMQImpl.readContentHeaderFrom(f.getInputStream());
-            this.remainingBodyBytes = this.contentHeader.getBodySize();
+            long bodySize = this.contentHeader.getBodySize();
+            if (bodySize >= this.maxBodyLength) {
+                throw new IllegalStateException(format(
+                    "Message body is too large (%d), maximum size is %d",
```
michaelklishin, line 110: Should we make this say "maximum configured size is …" and perhaps "See ConnectionFactory#setMaxInboundMessageBodySize if you need to increase the limit.".

PR #1096, opened 2023-08-23
chore: bump SLF4j to 2.0.7 (current)

milgner:
## Proposed Changes

Upgrade the dependency on SLF4j to the current 2.0 version. Interfaces are compatible but have been enhanced.

(I was struggling with two different interface versions on my classpath after adding the RabbitMQ client to a project which already uses SLF4J 2.0)

## Types of Changes

- [x] Bugfix (non-breaking change which ~~fixes~~ contributes to issue #1066)

## Checklist

- [ ] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)

## Further Comments

Did not sign the CA as this is a trivial change that is excluded in `CONTRIBUTING.md` - please let me know if it's still necessary.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
 
     <spotless.check.skip>true</spotless.check.skip>
-    <slf4j.version>1.7.36</slf4j.version>
+    <slf4j.version>2.0.7</slf4j.version>
     <metrics.version>4.2.19</metrics.version>
     <micrometer.version>1.11.3</micrometer.version>
     <micrometer-tracing.version>1.1.4</micrometer-tracing.version>


michaelklishin:
@milgner thank you!

There was an incompatibility at some point in the 2.x series. I no longer remember what it was but it did result in CI failures. Let's see.

milgner:
> @milgner thank you!
> 
> There was an incompatibility at some point in the 2.x series. I no longer remember what it was but it did result in CI failures. Let's see.

Fingers crossed then... I ran the "essential" tests on my local system as described in the README and those passed without a problem. I assume there are more running here - let's hope these run now, too.

michaelklishin:
@milgner seems to be good with `2.0.7`. Let's wait to see if @acogoluegnes has any objections.

acogoluegnes:
This is a breaking change. SLF4J 1.x and 2.x  bindings are not compatible. So an application using a 1.x binding would break as soon as it upgraded to the Java client that uses SLF4J 2.x.

SLF4J 1.x and 2.x are binary-compatible though: applications using the Java client can override  SLF4J to 2.x and use the appropriate binding. This is what PerfTest does.

Unless I misunderstood how SLF4J 2.x behaves, we'll have to stick to 1.x to avoid breaking applications, especially considering they can override the dependency.

milgner:
When I tried to override the dependency, I got an error about incompatible interfaces :thinking: Unfortunately I don't have it at hand anymore - I'll try and override once more.

acogoluegnes:
Consider PerfTest a regular application that uses the Java client library, it depends on the Java client library and uses SLF4J 2.x:

https://github.com/rabbitmq/rabbitmq-perf-test/blob/b10757bcf4b32e7fb16912fc2316ede70f817f5f/pom.xml#L58-L59
https://github.com/rabbitmq/rabbitmq-perf-test/blob/b10757bcf4b32e7fb16912fc2316ede70f817f5f/pom.xml#L114-L123

It then uses Logback as an implementation (the binding is included in logback):

https://github.com/rabbitmq/rabbitmq-perf-test/blob/b10757bcf4b32e7fb16912fc2316ede70f817f5f/pom.xml#L67
https://github.com/rabbitmq/rabbitmq-perf-test/blob/b10757bcf4b32e7fb16912fc2316ede70f817f5f/pom.xml#L160-L177

PerfTest uses logback 1.3.x because it supports Java 8.

See https://www.slf4j.org/faq.html#changesInVersion200.

PR #1493, opened 2024-12-02
Handle CancelledKeyException in NIO loop

acogoluegnes:
This is a connection exception, we don't want to stop the NIO loop for it.


NicolasGeraud:
hi @acogoluegnes when do you expect to release milestone 5.24.0?

Thanks for this quick fix!

acogoluegnes:
I was expecting some feedback from the original reporter, I think he is a colleague of yours. If you can confirm you tested the snapshot and the bug did not occur I can cut a release.

NicolasGeraud:
sure I ask him, thanks Arnaud.

acogoluegnes:
[Original ML thread](https://groups.google.com/g/rabbitmq-users/c/R7sI3P32kdE/m/S9faAySSAAAJ).

PR #1824, opened 2025-12-08
Add ConfirmationPublisher for async publisher confirmations

lukebakken:
Traditional publisher confirms in the Java client require manual tracking of sequence numbers and manual correlation of `basic.return` messages. This makes per-message error handling complex and provides no built-in async pattern, backpressure mechanism, or message correlation support.

This change introduces `ConfirmationPublisher`, a standalone `AutoCloseable` that provides automatic publisher confirmation tracking with a `CompletableFuture`-based API, optional bounded-outstanding backpressure, and a generic context parameter for message correlation. It owns a dedicated channel created from the supplied `Connection` and requires no modifications to `Channel`, `ChannelN`, or any other existing class.

## API

```java
// Unlimited outstanding confirmations
ConfirmationPublisher publisher = ConfirmationPublisher.create(connection);

// At most 100 unconfirmed messages; basicPublishAsync blocks at the limit
ConfirmationPublisher publisher = ConfirmationPublisher.create(connection, 100);

// Custom executor for future completion
ConfirmationPublisher publisher = ConfirmationPublisher.create(connection, 100, executor);

CompletableFuture<T> future =
    publisher.basicPublishAsync(exchange, routingKey, props, body, context);
```

The future completes with the caller-supplied context on `basic.ack`, or exceptionally with a `PublishException` on `basic.nack` or `basic.return`.

## Design

The publisher owns a private channel, so no external publish can desynchronize the delivery-tag space. It keys its outstanding map by the channel's authoritative `getNextPublishSeqNo()`, read atomically with the publish under a lock, rather than maintaining a shadow counter. Futures are completed on a configurable executor (default `ForkJoinPool.commonPool()`), never on the connection's I/O thread.

A `basic.return` frame carries no delivery tag, so an `x-seq-no` header is injected only when `mandatory = true` (returns are impossible otherwise) to correlate the return back to a publish. Non-mandatory publishes avoid the header entirely. Note that when a mandatory message is routable, the broker echoes this header on delivery, so its consumers observe it; applications that must not expose it should publish such messages through their own channel.

## Recovery and failure handling

- When the connection supports automatic recovery, the publisher keeps working across recoveries. A `RecoveryListener` fails all outstanding futures at the start of recovery, before `confirm.select` resets the delivery-tag sequence, so a reused tag cannot mis-correlate against a stale entry.
- A connection-level publish failure is left to the natural shutdown and recovery path. A publish failure that leaves the channel genuinely stuck open with an advanced sequence number aborts the private channel, since correlation could no longer be trusted.
- Publishing while a channel is transiently not in confirm mode (mid-recovery) fails the future rather than sending an untracked message.
- Completions fall back to the calling thread if the configured executor rejects them, so a saturated or shut-down user executor never leaves a future hanging.

## PublishException

`PublishException` carries the publish sequence number, the user context, and, for a `basic.return`, the full `Return` (reply code and text, exchange, routing key, properties, body). Use `isReturn()` to distinguish a return from a nack; `getReturned()` is null for a nack.


src/main/java/com/rabbitmq/client/impl/ChannelN.java
```diff
@@ -90,6 +90,19 @@ public class ChannelN extends AMQChannel implements com.rabbitmq.client.Channel
     /** Whether any nacks have been received since the last waitForConfirms(). */
     private volatile boolean onlyAcksReceived = true;
 
+    /** Publisher confirmation tracking enabled */
+    private boolean publisherConfirmationTrackingEnabled = false;
+
+    /** Maximum outstanding confirmations (0 = unlimited) */
+    private int maxOutstandingConfirms = 0;
```
acogoluegnes: Should be final.

src/main/java/com/rabbitmq/client/ChannelOptions.java
```diff
@@ -0,0 +1,123 @@
...
+     * @param publisherConfirmationsEnabled enable publisher confirmations on the channel
+     *        (calls {@link Channel#confirmSelect()} automatically)
+     * @param publisherConfirmationTrackingEnabled enable automatic tracking and awaiting of confirmations
+     *        via {@link Channel#basicPublishAsync} methods
+     * @param maxOutstandingConfirms maximum number of unconfirmed messages allowed before
+     *        {@link Channel#basicPublishAsync} blocks. Set to 0 for unlimited (default).
+     *        This prevents unbounded memory growth when publishing faster than the broker can confirm.
+     * @throws IllegalArgumentException if tracking is enabled but confirmations are not enabled
+     */
+    public ChannelOptions(boolean publisherConfirmationsEnabled,
```
acogoluegnes: I'd suggest a builder-based approach instead of a constructor-based one. The builder has reasonable defaults, it provides a fluent API with short method names and ultimately calls a private constructor.

If we add a new setting in the future, we'll have to create a 4-parameter constructor and keep the 3-parameter one for backward compatibility. We avoid this with the builder approach.

acogoluegnes:
Thanks for this contributon @lukebakken. I think it is on the right track. I added some comments in the code. Just a few more remarks:

* the new methods in the public interfaces should have a default implementation if we back port the PR to 5.x. This is for backward compatibility. Throwing an exception is good enough.
* synchronized blocks and Object-based synchronization is considered old-style Java concurrency. We could see if there are more modern utilities in the Java concurrency toolkit to handle what we need in `ChannelN`. I'm not super opinionated on this topic though and this is an implementation details we can polish later.
* the bulk of the PR is in `ChannelN`, it is not rocket science but add some non-trivial logic to a somewhat already complex class. I was wondering if it could be possible to externalize all the logic in a `PublishConfirmState` class used from `ChannelN`. This state class would register regular return and confirm listeners on the channel. This could even be an interface with a no-op implementation when the feature is not activated, minimalizing the impact on `ChannelN`. We can discuss this design when all the features are added.

src/main/java/com/rabbitmq/client/ChannelOptions.java
```diff
@@ -0,0 +1,123 @@
...
+ */
+public class ChannelOptions {
+    /**
+     * Header name for tracking publish sequence numbers in message properties.
+     * <p>
+     * When publisher confirmation tracking is enabled, this header is automatically
+     * added to each published message with the message's sequence number as the value.
+     * This allows the library to correlate Basic.Return responses with the correct message.
+     */
+    public static final String PUBLISH_SEQUENCE_NUMBER_HEADER = "x-java-pub-seq-no";
```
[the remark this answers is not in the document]
lukebakken: We can use x-seq-no or something else? `x-java-pub-seq-no` mirrors what was added in the .NET client.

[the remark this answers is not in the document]
michaelklishin: `x-seq-no` sounds reasonable to me.

lukebakken:
Alrighty gang, all set I think. Small tutorial here - https://github.com/rabbitmq/rabbitmq-tutorials/pull/707

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
```diff
@@ -0,0 +1,934 @@
...
+        try
+        {
+            // Acquire rate limiter permit if configured
+            if (rateLimiter != null)
+            {
+                permit = rateLimiter.acquire();
+            }
+
+            // Get next sequence number
+            seqNo = nextSeqNo.getAndIncrement();
```
lukebakken: **F1: confirms can complete the wrong futures under concurrent publishes**

`nextSeqNo.getAndIncrement()` and `delegate.basicPublish(...)` are not atomic with each other, and `handleAck`/`handleNack` look up the broker's delivery tag in a map keyed by this wrapper counter. The two only align if increment order matches wire order.

Failure: thread A takes seqNo 1, thread B takes seqNo 2, but B's `basicPublish` reaches the wire first, so the broker assigns delivery tag 1 to B's message. A nack of tag 1 (B's message) completes A's future exceptionally and B's future successfully - exactly inverted. This contradicts the interface javadoc's claim that the method is thread-safe.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
```diff
@@ -0,0 +1,934 @@
...
+        }
+
+        this.delegate = delegate;
+        this.rateLimiter = rateLimiter;
+
+        int initialCapacity = (rateLimiter != null) ? rateLimiter.getMaxConcurrency() : 16;
+        this.confirmations = new ConcurrentHashMap<>(initialCapacity > 0 ? initialCapacity : 16);
+
+        // Enable publisher confirmations on the delegate channel
+        delegate.confirmSelect();
```
lukebakken: **F2: sequence spaces permanently desynchronize in several reachable states**

The wrapper assumes the delegate's delivery tags start at 1 in lockstep with its own counter. That breaks when:

- the channel already published in confirm mode before wrapping: `ChannelN.confirmSelect()` short-circuits via `confirmSelectActivated` and only initializes `nextPublishSeqNo` when it is 0, so the broker's next tag is N+1 while the wrapper starts at 1
- anyone publishes directly on the delegate reference the caller still holds
- a publish throws after `ChannelN.basicPublish` consumed its tag (it increments before `transmit`)

In each case `confirmations.remove(tag)` finds nothing (or the wrong entry): futures hang forever or complete for the wrong message, and rate-limiter permits leak.

lukebakken (commented):
Code review findings, tracked as inline comments so each can be resolved as it is addressed. Ranked roughly by severity in each comment's first line (F1 = most severe, F10 = least).

Note that F1, F2, F3, and part of F10 share one root cause: the wrapper keeps its own `AtomicLong` sequence counter parallel to the broker's delivery-tag space instead of using the channel's authoritative `getNextPublishSeqNo()` (or integrating tracking into `ChannelN` itself, where the counter and `unconfirmedSet` already live). Fixing that design point addresses all four together.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
```diff
@@ -0,0 +1,934 @@
...
+ * The {@link #basicPublish} methods are not supported and will throw
+ * {@link UnsupportedOperationException}. Use {@link #basicPublishAsync} instead.
+ */
+public class ConfirmationChannelN implements ConfirmationChannel
+{
+    private static final String SEQUENCE_NUMBER_HEADER = "x-seq-no";
+
+    private final Channel delegate;
+    private final RateLimiter rateLimiter;
+    private final AtomicLong nextSeqNo = new AtomicLong(1);
```
lukebakken: **F3: permanently hangs after connection recovery**

When wrapping an `AutorecoveringChannel`, recovery re-adds the wrapper's listeners and re-issues `confirmSelect()` on a fresh channel whose delivery tags restart at 1 - but `nextSeqNo` is never reset (no recovery hook, no documented limitation). Post-recovery publishes are stored under keys continuing from pre-recovery values (101, 102, ...) while acks arrive tagged 1, 2, 3; nothing ever matches, every post-recovery future hangs, and with a `RateLimiter` configured the permits are never released, deadlocking publishers in `acquire()`. There is no test covering recovery.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
```diff
@@ -0,0 +1,934 @@
...
+                }
+                entry.releasePermit();
+            }
+        }
+    }
+
+    private void handleReturn(int replyCode, String replyText, String exchange,
+                               String routingKey, BasicProperties props, byte[] body)
+    {
+        Object seqNumObj = props.getHeaders().get(SEQUENCE_NUMBER_HEADER);
```
lukebakken: **F4: NPE on returned messages without headers kills the channel**

`props.getHeaders()` is dereferenced without a null check. Any mandatory publish on the raw delegate (which the caller still holds in the documented usage) with e.g. `MessageProperties.MINIMAL_BASIC` produces a `basic.return` with null headers: `handleReturn` throws NPE, `ChannelN.callReturnListeners` catches it outside the listener loop (skipping remaining user `ReturnListener`s), and the default `StrictExceptionHandler`-derived handler closes the channel, aborting every in-flight confirmation.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
```diff
@@ -0,0 +1,934 @@
...
+    }
+
+    @Override
+    public boolean removeConfirmListener(ConfirmListener listener)
+    {
+        return delegate.removeConfirmListener(listener);
+    }
+
+    @Override
+    public void clearConfirmListeners()
```
lukebakken: **F5: `clearConfirmListeners()`/`clearReturnListeners()` silently destroy tracking**

Both are delegated verbatim, so they also remove the internal listeners the constructor installed. After a user clears their own listeners (a normal `Channel` operation), publishes still succeed but acks never reach `handleAckNack`: every pending and future `basicPublishAsync` future hangs forever and permits leak, with no error surfaced. The delegating overrides need to filter out (or re-install) the internal listeners, or tracking needs to live where the public listener API cannot clobber it.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
```diff
@@ -0,0 +1,934 @@
...
+            {
+                props = MessageProperties.MINIMAL_BASIC;
+            }
+            props = addSequenceNumberHeader(props, seqNo);
+
+            // Publish to delegate channel
+            // Note: Metrics are collected by the delegate's MetricsCollector
+            delegate.basicPublish(exchange, routingKey, mandatory, props, body);
+        }
+        catch (IOException | AlreadyClosedException | InterruptedException e)
```
lukebakken: **F6: cleanup catch is too narrow, and the interrupt flag is swallowed**

Two issues:

- Only `IOException | AlreadyClosedException | InterruptedException` trigger cleanup. A raw `ShutdownSignalException` (the parent of `AlreadyClosedException`, thrown during connection-level shutdown) or any other unchecked exception escapes after the entry was inserted and the permit acquired, stranding both forever; after `maxConcurrency` such failures all publishers deadlock in `acquire()`. A `finally`-based cleanup guarded by a success flag (or `catch (Exception)`) closes the gap
- The `InterruptedException` branch never calls `Thread.currentThread().interrupt()`, so the interrupt is silently swallowed and cooperative cancellation breaks

src/main/java/com/rabbitmq/client/ThrottlingRateLimiter.java
```diff
@@ -0,0 +1,156 @@
...
+
+        private Permit(ThrottlingRateLimiter limiter) {
+            this.limiter = limiter;
+        }
+
+        /**
+         * Releases this permit, returning it to the rate limiter pool.
+         */
+        public void release() {
+            if (!released) {
```
lukebakken: **F7: permit double-release breaks the concurrency cap permanently**

`released` is a plain non-volatile boolean and `release()` is an unsynchronized check-then-act. `ConfirmationChannelN.handleShutdown` iterates `confirmations.values()` and releases permits without removing entries (the `clear()` happens after the loop), while the connection main-loop thread can concurrently process a final ack for the same entry and also release. Both threads observe `released == false` and both call `semaphore.release()`: a `ThrottlingRateLimiter(10)` now holds 11 permits for the rest of its lifetime. `released` should be an `AtomicBoolean` with `compareAndSet`, and the shutdown path should claim entries via `remove()` like the ack path does.

src/main/java/com/rabbitmq/client/ConfirmationChannel.java
```diff
@@ -0,0 +1,149 @@
...
+ *         }
+ *         return null;
+ *     });
+ * }</pre>
+ *
+ * @see Channel
+ * @see PublishException
+ * @see com.rabbitmq.client.impl.ConfirmationChannelN
+ */
+public interface ConfirmationChannel extends Channel
```
lukebakken: **F8: extending `Channel` while throwing `UnsupportedOperationException` from core methods breaks every `Channel`-typed caller**

`ConfirmationChannelN` throws from all three `basicPublish` overloads and all four `waitForConfirms*` methods. Anything typed against `Channel` accepts a `ConfirmationChannel` at compile time and explodes at runtime - including this library's own `RpcClient` (`_channel.basicPublish(...)` at RpcClient.java:212) and `RpcServer`. The interface javadoc does not document the throwing behavior. Either `basicPublishAsync` belongs on `ChannelN` itself (where confirm state already lives and all `Channel` methods keep working), or `ConfirmationChannel` should be a small standalone type that composes a `Channel` without claiming to be one.

src/main/java/com/rabbitmq/client/ThrottlingRateLimiter.java
```diff
@@ -0,0 +1,156 @@
...
+     * Acquires a permit, blocking if necessary until one is available.
+     * Applies throttling delay if the number of available permits falls below the threshold.
+     *
+     * @return A permit that must be released when the operation completes
+     * @throws InterruptedException if the thread is interrupted while waiting
+     */
+    public Permit acquire() throws InterruptedException {
+        int delay = calculateDelay();
+        if (delay > 0) {
+            Thread.sleep(delay);
```
lukebakken: **F9: unconditional `Thread.sleep` of up to ~1000 ms per publish**

`calculateDelay()` returns `(1.0 - available/max) * 1000` ms once available permits drop below the 50% threshold, computed once from a stale snapshot before `semaphore.acquire()` and never re-evaluated as permits free up. Just past 50% utilization every `basicPublishAsync` sleeps ~500 ms even when permits are free, capping a publisher thread at ~2 msg/s regardless of broker speed; called from a consumer callback it stalls that work-service thread. The blocking `semaphore.acquire()` already provides backpressure on its own; if a ramp is desired it should be event-driven rather than a fixed sleep.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
```diff
@@ -0,0 +1,934 @@
...
+    private void handleNack(long seqNo, boolean multiple)
+    {
+        handleAckNack(seqNo, multiple, true);
+    }
+
+    private void handleAckNack(long seqNo, boolean multiple, boolean nack)
+    {
+        if (multiple)
+        {
+            for (Long seq : new ArrayList<>(confirmations.keySet()))
```
lukebakken: **F10: O(n) scan per multiple-ack, executed on the connection's frame-reader thread, with inline future completion**

Confirm listeners run synchronously on the connection main loop (`MainLoop.run -> handleFrame -> processAsync -> callConfirmListeners`, with no work-pool indirection). This branch copies the entire `confirmations.keySet()` into an `ArrayList` and scans it on every `multiple=true` confirm - O(n) per ack, O(n^2) over a burst with 10k outstanding publishes - and each `future.complete()` then runs non-async user continuations inline on that same thread, stalling deliveries, heartbeats, and every other channel on the connection. `ChannelN` solves the same problem with `unconfirmedSet.headSet(seqNo + 1).clear()` on a sorted structure; a `ConcurrentSkipListMap.headMap(seqNo, true)` drain plus completing futures via an executor is the matching fix here.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
lukebakken: Addressed in the reimplementation. The tracking no longer uses a shadow `AtomicLong` counter. `ConfirmationPublisher` owns a private channel and, under a `ReentrantLock`, reads `channel.getNextPublishSeqNo()` and calls `basicPublish` as one atomic step. The map key is therefore always the broker's delivery tag, so the increment order cannot diverge from wire order.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
lukebakken: Root cause removed. The publisher owns a private channel it creates and puts into confirm mode itself, so there are no pre-wrapping publishes and no external publishes on the delegate to desynchronize the tag space. It keys by the channel's authoritative `getNextPublishSeqNo()` rather than a parallel counter. The publish-after-tag-consumed case is handled explicitly: a non-`IOException` failure that leaves the channel stuck open aborts it, while an `IOException` is left to the shutdown and recovery path (which resets the counter).

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
lukebakken: Fixed. Keying by the channel's own `getNextPublishSeqNo()` means the counter resets to 1 on recovery along with the broker's tags (verified experimentally and now covered by a `publisherSurvivesConnectionRecovery` test). In addition, a `RecoveryListener` fails all outstanding futures at the start of recovery, before `confirm.select` resets the tag space, so a reused tag cannot mis-correlate against a stale entry, and permits are released on that path.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
lukebakken: Fixed. `extractSequenceNumber` is null-safe: it returns null when the properties or headers are null, or when the header value is not a `Long`. A `basic.return` without a usable sequence-number header is ignored and logged at debug rather than throwing.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
lukebakken: No longer possible. The publisher owns a private channel that the application never holds a reference to, so user code cannot call `clearConfirmListeners()` / `clearReturnListeners()` on it and clobber the internal tracking listeners.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
lukebakken: Both addressed. The publish-failure cleanup now runs from a `catch (Exception)` block that removes the entry and releases the permit, so no unchecked exception (including a raw `ShutdownSignalException`) can strand an entry. The `InterruptedException` path during permit acquisition now restores the interrupt flag with `Thread.currentThread().interrupt()`.

src/main/java/com/rabbitmq/client/ThrottlingRateLimiter.java
lukebakken: Fixed. Permit release uses `AtomicBoolean.compareAndSet` so a slot is released at most once. The shutdown and recovery drains claim each entry via `pollFirstEntry()` (a remove) before completing it, so the ack path and the drain can never both act on the same entry. The `ThrottlingRateLimiter` itself is gone (see F9).

src/main/java/com/rabbitmq/client/ConfirmationChannel.java
lukebakken: Fixed by the structural change this review recommended. `ConfirmationPublisher` is a standalone `AutoCloseable` that composes a `Channel` rather than extending it, so no `Channel`-typed caller (including this library's `RpcClient` / `RpcServer`) can receive it and fail at runtime. There are no `UnsupportedOperationException`-throwing overrides.

src/main/java/com/rabbitmq/client/ThrottlingRateLimiter.java
lukebakken: Fixed. The `ThrottlingRateLimiter` and its fixed `Thread.sleep` are removed entirely. Backpressure is now a plain `Semaphore(maxOutstandingConfirms)`: `basicPublishAsync` blocks only when the outstanding limit is actually reached, with no per-publish sleep. An unbounded mode (no semaphore) is also available.

src/main/java/com/rabbitmq/client/impl/ConfirmationChannelN.java
lukebakken: Fixed. Multiple-acks now drain `outstanding.headMap(deliveryTag, true)` in a single pass and clear the view, matching `ChannelN`'s own `headSet(...).clear()` pattern. Futures are completed via a configurable executor (default `ForkJoinPool.commonPool()`), so user continuations never run inline on the connection's frame-reader thread.

sdewhitt:
This looks really handy!

PR #1978, opened 2026-06-01
fix vHost not set to default when trailing slash in URI

ahmed-debbech:
## Proposed Changes

This Pull Request addresses a [bug](https://github.com/rabbitmq/rabbitmq-java-client/discussions/1977) when using URI to create a connection to RabbitMq Server and that URI has a trailing slash indicating virtual host `/`, `setUri()` from `ConnectionFactory` fails to parse that and sets virtual host to empty string.
So this Pull request adds the check when the URI has a trailing slash only it sets the virtual host to the default value else it sets the provided value. (example URI `amqp://guest:guest@localhost:5672/`)

## Types of Changes

- [x] Bugfix (non-breaking change which fixes issue #1977)

## Checklist

- [ ] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)

### DIFF
--- a/src/main/java/com/rabbitmq/client/ConnectionFactory.java
+++ b/src/main/java/com/rabbitmq/client/ConnectionFactory.java
@@ -389,7 +389,12 @@ public class ConnectionFactory implements Cloneable {
         throw new IllegalArgumentException("Multiple segments in " + "path of AMQP URI: " + path);
       }
 
-      setVirtualHost(uriDecode(uri.getPath().substring(1)));
+      String pathDecoded = uriDecode(uri.getPath().substring(1));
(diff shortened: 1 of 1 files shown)


michaelklishin:
The [Examples](https://www.rabbitmq.com/docs/uri-spec#appendix-a-examples) section in the AMQP URI explains clearly, just like Luke did in https://github.com/rabbitmq/rabbitmq-java-client/discussions/1977, that the trailing slash, if present, is considered to be a separator, so the virtual host value extracted will be an empty string.

You are not fixing "a bug" here, you are making the client work the way you want but not the way the URI parsing spec describes it. The current behavior has been around for well over a decade, and is consistent across the clients our team maintains.

PR #1994, opened 2026-06-23
Stricter validation of pre-negotiated values in the frame reader

michaelklishin:


michaelklishin:
@mergifyio backport v5.x

PR #1996, opened 2026-06-23
Enforce inbound frame max more strictly

acogoluegnes:
Follow-up of #1994


acogoluegnes:
@Mergifyio backport v5.x

PR #2013, opened 2026-07-08
Reject negative body size when parsing content headers

acogoluegnes:
consumeHeaderFrame read the content header's body-size field as a signed long and only checked it against the configured maximum with `bodySize >= maxBodyLength`. A wire value with the high bit set decodes to a negative long, which always satisfies that comparison, bypassing the size check. The negative remainingBodyBytes then made the command complete with an empty body instead of throwing, so any body frames that followed were parsed as the start of a new command.

Reject negative body sizes the same way oversized ones are already rejected.

Add BrokenFramesTest#negativeBodySizeTriggersConnectionClosure.


acogoluegnes:
@Mergifyio backport v5.x

PR #2015, opened 2026-07-08
Honor unlimited negotiated frame_max when capping inbound message size

acogoluegnes:


acogoluegnes:
@Mergifyio backport v5.x

PR #2064, opened 2026-08-27
Be more defensive when parsing shortstr values in tables

michaelklishin:
Note: this enforces the AMQP 0-9-1 255 byte limit on short strings.

We can consider bumping it a bit to reduce the risk of this being a minor breaking change.
In any case, the 255 byte limit has been around for as long as the protocol and RabbitMQ has (19-20 years depending on how you count), just not necessarily universally enforced.


michaelklishin:
@mergifyio backport `v5.x`

michaelklishin:
@mergifyio backport v5.x

